PR #731, opened 2023-11-03
Set royalty limit to 6 when creating a collection

JasonTulp:
As per [TRN-256](https://futureverse.atlassian.net/browse/TRN-256)

Royalties can currently be set with 8 entitlements which will then fail when the user tries to list the token. This is because when the token is listed some extra entitlements are added for network fee and marketplace fee. This change reduces the limit of royalties for NFT and SFT collections to MAX_ENTITLEMENTS - 2 so that when royalties are added, it leaves room for these two entitlements at list time. Note we went for this approach to avoid any unnecessary storage migrations. Realistically, most use cases will only involve one or two entitlements so lowering the limit to 6 shouldn't affect any external teams

[TRN-256]: https://futureverse.atlassian.net/browse/TRN-256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zees-dev (approved):
LGTM 👍

PR #743, opened 2023-11-22
Cleanup mocks and reduce repeated test code

JasonTulp:
- Adds some common test utilities to pallet/common/test_utils
- Adds test_prelude to help simplify imports in mocks and tests
- Adds common TestExternalities builder in seed_primitives/test_utils
   - Note: some pallets require their own GenesisConfig or custom setup so have been left as is
- Removes has_event functions in favor of System::assert_last_event
- Replaces mock pallet config with macros where possible


pallet/assets-ext/src/mock.rs
```diff
@@ -298,20 +143,11 @@ pub fn test_ext() -> TestExt {
 struct AssetsFixture {
```
zees-dev, line 143: tip: I'd reckon there's an oppurtunity here to remove this; to decrease the discrepancies between the mocks

pallet/assets-ext/src/impls.rs
```diff
@@ -352,10 +352,9 @@ where
 
 #[cfg(test)]
 mod tests {
```
zees-dev, line 354: nit: maybe we should move these tests to the `tests` file?

pallet/assets-ext/src/mock.rs
```diff
@@ -14,190 +14,35 @@
...
-		AssetsExt: pallet_assets_ext::{Pallet, Call, Storage, Event<T>},
-		EVM: pallet_evm::{Pallet, Config, Call, Storage, Event<T>},
-		TimestampPallet: pallet_timestamp::{Pallet, Call, Storage, Inherent},
+		System: frame_system,
+		Balances: pallet_balances,
+		Assets: pallet_assets,
+		AssetsExt: pallet_assets_ext,
+		EVM: pallet_evm,
+		TimestampPallet: pallet_timestamp,
+		FeeControl: pallet_fee_control,
```
zees-dev, line 36: This pallet did not exist before; should try to keep the runtime minimal (pull in only pallets which are required for the tests).
Is there any reason why we added this?

pallet/common/src/lib.rs
```diff
@@ -34,7 +34,8 @@ use seed_primitives::{
 	AssetId, Balance, CollectionUuid, MetadataScheme, SerialNumber, TokenId,
 };
 
-pub mod mock;
+pub mod test_utils;
+pub use test_utils::test_prelude;
```
zees-dev, line 38: not needed for in production code; we should put this behind the `test` feature flag (so it's not compiled in the production binary).

pallet/common/Cargo.toml
```diff
@@ -16,6 +16,7 @@ scale-info = { version = "2.3.0", default-features = false, features = ["derive"
 
 # Substrate
 frame-support = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.30" }
+frame-system = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.30" }
```
zees-dev, line 19: we could move these to dev-dependencies now (if we put all test utils under the `test` feature/compiler flag)

pallet/echo/src/tests.rs
```diff
@@ -15,21 +15,16 @@
...
-use sp_core::H160;
+use crate::mock::{test_storage::NextEventProofId, Echo, MockEthereumEventRouter, System, Test};
+use frame_support::storage::StorageValue;
+use seed_pallet_common::test_prelude::*;
 use sp_runtime::traits::AccountIdConversion;
 
 #[test]
 fn ping_works_from_runtime() {
-	ExtBuilder::default().build().execute_with(|| {
+	test_ext::<Test>().build().execute_with(|| {
```
zees-dev, line 25: Hmm not sure if i agree with having this helper - since it appears to be an unnecessary abstraction over `TestExt::default()` (or equivalent).
Would simply use that instead of introducing another indirection.

pallet/token-approvals/src/benchmarking.rs
```diff
@@ -24,7 +24,7 @@ use frame_support::assert_ok;
 use frame_system::RawOrigin;
 use seed_primitives::{nft::OriginChain, MetadataScheme};
 
-use crate::Pallet as TokeApprovals;
+use crate::Pallet as TokenApprovals;
```
zees-dev, line 27: lol, nice catch

primitives/src/lib.rs
```diff
@@ -23,6 +23,8 @@ pub use types::*;
 pub mod ethy;
 pub mod nft;
 mod signature;
+pub mod test_utils;
```
zees-dev, line 26: again lets put this behind the `test` feature/compiler flag

primitives/src/test_utils.rs
```diff
@@ -0,0 +1,109 @@
...
+impl<T> TestExt<T>
+where
+	T: frame_system::Config + pallet_balances::Config + pallet_assets::Config,
+	<T as pallet_balances::Config>::Balance: From<Balance>,
+{
+	/// Create new TestExt
+	pub fn new() -> Self {
+		Self { balances: vec![], xrp_balances: vec![], block_number: 1 }
+	}
+}
```
zees-dev, line 26: We use `::default()` in some places too; do you think we should just implement the `default` trait for this and use it instead of `new()`?

primitives/src/test_utils.rs
```diff
@@ -0,0 +1,109 @@
...
+
+/// Small helper to create a TestExt
+pub fn test_ext<T>() -> TestExt<T>
+where
+	T: frame_system::Config + pallet_balances::Config + pallet_assets::Config,
+	<T as pallet_balances::Config>::Balance: From<Balance>,
+{
+	TestExt::<T>::new()
+}
```
zees-dev, line 35: Don't this this function is needed, simple enough to just call `new` or `default()`

primitives/Cargo.toml
```diff
@@ -26,6 +26,10 @@ sp-core = { git = "https://github.com/paritytech/substrate", default-features =
 sp-io = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.30" }
 sp-runtime = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.30" }
 sp-std = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.30" }
+frame-support = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.30" }
+frame-system = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.30" }
+pallet-balances = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.30" }
+pallet-assets = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.30" }
```
zees-dev, line 32: will need to go under dev-deps once we put test_utils under the `test` compiler flag

primitives/Cargo.toml
```diff
@@ -39,6 +43,10 @@ default = ["std"]
 std = [
     "hex/std",
     "codec/std",
+    "frame-support/std",
+    "frame-system/std",
+    "pallet-balances/std",
+    "pallet-assets/std",
```
zees-dev, line 49: these changes won't be needed then either

pallet/vortex-distribution/src/mock.rs
```diff
@@ -199,6 +180,7 @@ impl TestExt {
 		self
 	}
 
+	#[allow(dead_code)]
```
zees-dev, line 183: maybe requires the `try-runtime` flag instead?
If it's not used there either, then worthwhile simply removing completely unused code

zees-dev (commented):
Awesome work! 🚀
A lot of great quality of life changes in this PR which will make testing more consistent and easier 👍

pallet/assets-ext/src/mock.rs
JasonTulp: I wanted to remove this one as well, however due to the AssetsExt mock calling the GenesisConfig in the build() function it meant that the generic mock did not work (And caused some tests to fail) We could include AssetsExt as a strict dependency for the generic mock, but this would mean all mocks would need assetsExt as well. (Maybe not a big deal?) What are your thoughts?

pallet/assets-ext/src/impls.rs
JasonTulp: Hmm could do, I think it's a matter of taste. On one hand it's nice to have them in the same file as the structs themselves but on another it might be better to have all tests in the same file

pallet/assets-ext/src/mock.rs
JasonTulp: Yep! This was added due to the evm macro needing FeeControl internally. I figured it was better to use the macro and add the extra pallet then to leave it out and manually redefine the EVM pallet

pallet/common/src/lib.rs
JasonTulp: Ooh good catch, was meaning to change this over

pallet/echo/src/tests.rs
JasonTulp: Maybe I went a little overboard with trying to reduce the size of these tests haha, will remove the helper function

primitives/src/test_utils.rs
JasonTulp: I went with new() because it was smaller (slightly less characters) Probably more descriptive to use default though

pallet/assets-ext/src/mock.rs
zees-dev: Hmm I see, in that case imo leave as is. Would prefer not having AssetsExt as a hard dependency for mock runtimes 👍

JasonTulp: Oh actually I forgot the reason I didn't include AssetsExt is that we would end up with cyclic dependencies with assets-ext and seed-primitives. I had an idea to create a trait that could extend the TestExt with extra GenesisConfig but couldn't get it to work cleanly so decided to just keep the custom mock in AssetsExt

pallet/assets-ext/src/impls.rs
surangap: IMO it's okay to have a mix

pallet/common/src/lib.rs
```diff
@@ -34,8 +34,11 @@ use seed_primitives::{
 	AssetId, Balance, CollectionUuid, MetadataScheme, SerialNumber, TokenId,
 };
 
-pub mod mock;
+#[cfg(feature = "std")]
```
surangap, line 37: is this feature flag correct? should it be test?

JasonTulp: I tried many ways to get it to work with test but could not get it to work as intended. Ended up using the std flag instead (The same pattern is used for some Substrate test utils)

surangap (commented):
lgtm. thanks for working on this👏. Had a minor comment

surangap (approved):

PR #752, opened 2023-12-01
Update get state script.

surangap:
### Changelog

- Half per request key size to 1000 when value fetching
- Add error handling at value fetching

Note this PR will be effective combined with the other fix mentioned [here](https://futureverse.atlassian.net/browse/TRN-289?focusedCommentId=24773)


zees-dev (approved):
looks good; lets catch and handle these errors 👍

PR #754, opened 2023-12-05
remove extra transfer attempt of ROOT when redeem vortex from vault

leo-plugdefi:


pallet/vortex-distribution/src/lib.rs
```diff
@@ -710,19 +704,6 @@ pub mod pallet {
 					false,
 				)?;
```
zees-dev, line 705: wouldn't this code block also safe transfer root aswell?
Since AssetPrices can also include root tokens (according to the e2e test aswell)?

pallet/vortex-distribution/src/tests.rs
```diff
@@ -793,6 +803,109 @@ fn redeem_tokens_from_vault_should_work() {
...
+				),
+				1_000_000
+			);
+
+			run_to_block(end_block);
+
+			assert_ok!(Vortex::pay_unsigned(Origin::none(), vortex_dis_id, end_block));
+			assert!(
+				!System::events().iter().all(|record| {
+					println!("{:?}", record.event);
```
zees-dev, line 880: nit: remove print statements in tests

pallet/vortex-distribution/src/tests.rs
zees-dev, line 1: Can we also have a test to validate the double mint fails.
Idealy the same test should fail without the change, and pass with the change.

pallet/vortex-distribution/src/lib.rs
```diff
@@ -555,6 +555,7 @@ pub mod pallet {
 		///
 		/// `id` - The distribution id
 		#[pallet::weight(<T as pallet::Config>::WeightInfo::trigger_vtx_distribution())]
+		#[transactional]
```
JasonTulp, line 558: Why was transactional added?

pallet/vortex-distribution/src/tests.rs
leo-plugdefi: After implementing the necessary corrections, the current code logic cannot produce a failed case, any suggestions on how to do that?

pallet/vortex-distribution/src/lib.rs
leo-plugdefi: There are a few 'safe_transfer' operations triggered during the distribution of VTX, and if any of them fails, it could leave the system in a broken state. It might be more advisable to either ensure the success of all transactions or roll back in case of failure.

pallet/vortex-distribution/src/tests.rs
JasonTulp: It should include a test that passes with the change, but fails before the change was implemented (i.e. pass root in assets list and make sure it's only transfered once)

JasonTulp: Hey @leo-plugdefi Sorry yep that looks like it does the same thing, sorry was a bit hard to see as I would have expected to see the results of the test change but after pulling the code and looking through it, that makes sense 👍

JasonTulp (approved):

PR #758, opened 2023-12-07
Pallet XRPL

zees-dev:
## Context

This PR introduces the XRPL transaction pallet (`pallet-xrpl-transaction`) - which allows for the submission of signed XRPL transactions of the [`SignIn`](https://docs.xumm.dev/concepts/special-transaction-types) type by the XUMM mobile wallet to be submitted to the chain.

The signed transaction must contain specific memo data of the `extrinsic` - which encodes the `chain_id`, `nonce`, `max_block_number` and a hash (`blake256`) of the `scale_encoded_extrinsic` to be dispatched by the mapped ethereum address (from the provided pub key in the signed transaction).

The pallet only contains a single extrinsic `submit_encoded_xrpl_transaction` that is a self contained call - which means that the signed transaction is validated based on the data provided in the signed transaction itself.

A signature must be provided in the `submit_encoded_xrpl_transaction` call (2nd param) in addition to the actual call itself (to be which is hashed and signed in the memo data); these parameters are validated against the provided encoded/signed XRPL transaction.

[Original RFC](https://www.notion.so/futureverse/RFC-XUMM-wallet-integration-79cd2fd62e054cd49cf9398f07021ac5)

### Dependent PRs

- https://github.com/futureversecom/trn-frontier/pull/6
- https://github.com/futureversecom/trn-seed/pull/753
- https://github.com/futureversecom/xrpl-sdk-rust/pull/4


ken-centrality:
can we shorten the pallet to be just `xrpl` @zees-dev ?

zees-dev:
> can we shorten the pallet to be just `xrpl` @zees-dev ?

so, `pallet-xrpl`?
I do think it can create confusion with `xrpl-bridge` and mabye `xls-20` since it's not really like a top-level xrp crate.
What do you think?

p-dealwis:
@zees-dev I'm guessing because we are using the ethereum address scheme they can't hold assets on their address in TRN? And if so does that mean they will only be able to access TRN via their FPass account?

zees-dev:
> @zees-dev I'm guessing because we are using the ethereum address scheme they can't hold assets on their address in TRN? And if so does that mean they will only be able to access TRN via their FPass account?

Since the public key can be mapped to both XRP address scheme and TRN address scheme, we can use it to derive the ethereum address.
In fact this derived ethereum address is used to perform their futurepass address lookup.
Hence it is possible for both the mapped ethereum address and the respective futurepass address to hold assets.

pallet/xrpl/src/mock.rs
```diff
@@ -0,0 +1,179 @@
...
+// limitations under the License.
+// You may obtain a copy of the License at the root of this project source code
+
+use crate::{self as pallet_xrpl, *};
+use frame_support::{parameter_types, weights::WeightToFee, PalletId};
+use frame_system::EnsureRoot;
+use seed_pallet_common::test_prelude::*;
+use seed_primitives::{AccountId, Address, AssetId, Balance, Signature};
+use sp_core::H256;
+use sp_runtime::{generic, testing::Header, traits::LookupError};
```
JasonTulp, line 22: A lot of these imports are included within the test_prelude
```suggestion
use frame_support::weights::WeightToFee;
use seed_pallet_common::test_prelude::*;
use seed_primitives::{Address, Signature};
use sp_runtime::{generic, traits::LookupError};
```

pallet/xrpl/src/tests.rs
```diff
@@ -0,0 +1,218 @@
...
+				}));
+				assert_err!(
+					Executive::validate_transaction(TransactionSource::External, xt.into(), H256::default()),
+					TransactionValidityError::Invalid(InvalidTransaction::BadProof),
+				);
+
+				// validate account nonce is not incremented from any of the failure scanerios above
+				let tx = XRPLTransaction::try_from(tx_bytes.as_bytes_ref()).unwrap();
+				let caller: AccountId20 = tx.get_account().unwrap().into();
+				assert_eq!(System::account_nonce(&caller), 0);
```
JasonTulp, line 142: I don't think this check is useful as the nonce is only incrimented in pre_dispatch so won't be triggered with the validate_transaction calls even in a success case

pallet/xrpl/src/tests.rs
```diff
@@ -0,0 +1,218 @@
...
+					encoded_msg: BoundedVec::truncate_from(hex::decode("5916969036626990000000000000000000F236FD752B5E4C84810AB3D41A3C25807321033A0663EEAAD786F132CDBC25C7D2A6F8C14D55DB7B9AB52AFFB0D8A1C9A1010D81145B3DE9CEA3A77D69DD12F99C12E542907EE49E44F9EA7C0965787472696E7369637D30313A303A353A353030343030303134303464363937333633363836393635363632303464363136653631363736353634E1F1").unwrap()),
+					signature: BoundedVec::truncate_from(hex::decode("3045022100B877466D021B990299F5177E33AF2B2D4B40D2D01CF0889C26247BAEF7995C6F02207EEAD77A28D990F94C6F425EC134EAAD456282FD74CB09005656CA208E8A1476").unwrap()),
+					call: Box::new(call.clone()),
+				}));
+				assert_err!(
+					Executive::validate_transaction(TransactionSource::External, xt.into(), H256::default()),
+					TransactionValidityError::Invalid(InvalidTransaction::BadProof),
+				);
+
+				// encoded call for: chain_id = 0, nonce = 1, max_block_number = 5, extrinsic = System::remark; validates nonce too high
```
JasonTulp, line 81: I think this test is failing for the wrong reason. In my tests I couldn't get the `Executive::validate_transaction` to fail due to a too high nonce as it looks like it just triggeres `validate()` in check_nonce, not `pre_dispatch()` Which only checks if the nonce is stale (Not future) A better test would be to separate this part out, assert_ok on validation then assert_err on apply_extrinsic

pallet/xrpl/src/mock.rs
zees-dev: addressed: https://github.com/futureversecom/trn-seed/pull/758/commits/35a472cc9b964c12d07c40db75f56892b851bd3b

pallet/xrpl/src/tests.rs
zees-dev: addressed: https://github.com/futureversecom/trn-seed/pull/758/commits/5e1add01482f83b70e4b46c77a5e1ed973674145

pallet/xrpl/src/tests.rs
zees-dev: good catch!
Addressed: https://github.com/futureversecom/trn-seed/pull/758/commits/4163e93752d9c96bc2543671215b89d692b83ed8

p-dealwis:
> > @zees-dev I'm guessing because we are using the ethereum address scheme they can't hold assets on their address in TRN? And if so does that mean they will only be able to access TRN via their FPass account?
> 
> 
> 
> Since the public key can be mapped to both XRP address scheme and TRN address scheme, we can use it to derive the ethereum address.
> 
> In fact this derived ethereum address is used to perform their futurepass address lookup.
> 
> Hence it is possible for both the mapped ethereum address and the respective futurepass address to hold assets.
> 
> 

Perfect! So it seems like their wallet will be able to also sign Ethereum messages?

zees-dev:
> > > @zees-dev I'm guessing because we are using the ethereum address scheme they can't hold assets on their address in TRN? And if so does that mean they will only be able to access TRN via their FPass account?
> > 
> > 
> > Since the public key can be mapped to both XRP address scheme and TRN address scheme, we can use it to derive the ethereum address.
> > In fact this derived ethereum address is used to perform their futurepass address lookup.
> > Hence it is possible for both the mapped ethereum address and the respective futurepass address to hold assets.
> 
> Perfect! So it seems like their wallet will be able to also sign Ethereum messages?

The wallet allows for signing any arbitrary message in the memo data field.
This is how we pass in extrinsics - which are executed on the chain.
This is not an ethereum signed message however.

pallet/xrpl/src/lib.rs
```diff
@@ -0,0 +1,422 @@
+// Copyright 2022-2023 Futureverse Corporation Limited
```
surangap, line 1: might need to update this to cover 2024

pallet/xrpl/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+					e
+				})
+				.ok()?;
+			let ExtrinsicMemoData { chain_id, nonce, max_block_number, tip, hashed_call } = tx.get_extrinsic_data()
+				.map_err(|e| {
+					log::error!("⛔️ failed to extract extrinsic data from memo data: {:?}, err: {:?}", tx.memos, e);
+					e
+				})
+				.ok()?;
+			if chain_id != T::ChainId::get() {
```
surangap, line 154: This is regarding the extra validations that are not included in the apply path. https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/primitives/runtime/src/traits.rs#L1483
New self contained path is a bit different, but the idea here is that checks we perform on the validate and apply paths should be the same/similar. let's include all the checks in apply path as well.

pallet/xrpl/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+				return None;
+			}
+
+			let validations: XRPLValidations<T> = (
+				CheckNonZeroSender::new(),
+				CheckSpecVersion::<T>::new(),
+				CheckTxVersion::<T>::new(),
+				CheckGenesis::<T>::new(),
+				CheckEra::<T>::from(Era::immortal()),
+				CheckNonce::from(nonce.into()),
```
surangap, line 192: for futurepass calls, the nonce will be checked against the futurepass this way. Ideally it should be checked against the EOA account.

pallet/xrpl/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+
+	#[pallet::origin]
+	pub type Origin = seed_primitives::AccountId20;
+
+	#[pallet::pallet]
+	#[pallet::storage_version(STORAGE_VERSION)]
+	pub struct Pallet<T>(PhantomData<T>);
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config + pallet_transaction_payment::Config
```
surangap, line 288: wondering why we need `pallet_transaction_payment::Config` here

surangap (commented):
nice work👏. added couple of comments for further improvements/discussion

pallet/xrpl/src/lib.rs
zees-dev: I was under the assumption that we don't get into the apply path if the validate has failed.
Since the transactions will be dropped from the mempool; I think some of the unit tests also validate this (assuming that is the actual flow).

pallet/xrpl/src/lib.rs
zees-dev: Yes I believe this is the desired behaviour here - since the FP is making the transaction, not the EOA right

pallet/xrpl/src/lib.rs
zees-dev: Good catch!

pallet/xrpl/src/lib.rs
surangap: I think the mempool validations happens when a node receives a broadcasted transaction. But let's say block producer produces a block and that block gets broadcasted, when other nodes receive the block it will go block apply path. not sure if the mempool path is touched again.
Also there might be other ways(not the usual) to import blocks via subcommands like `ImportBlocks`. Hence better to do as mentioned in the framework.

pallet/xrpl/src/lib.rs
surangap: It should be the EOA nonce I believe. one of the main purpose of nonce is for replay protection. you can check the existing Futurepass as well. the sender nonce get checked and incremented, not the futurepass.

pallet/xrpl/src/lib.rs
zees-dev: in that case - would it not be better to have a function which is called by both validate and apply - for all self contained calls

pallet/xrpl/src/lib.rs
zees-dev: addressed: https://github.com/futureversecom/trn-seed/pull/758/commits/1db9590ab1a708cf28debc86c57520782ac61e88

pallet/xrpl/src/lib.rs
surangap: yeah, can do that as well. Some of the Signed extensions exactly do that. there might be cases where mempool and apply path validations might slightly differ. like nonce.

pallet/xrpl/src/lib.rs
zees-dev: addressed: https://github.com/futureversecom/trn-seed/pull/758/commits/efdcfddde304b832e3c789e0fa6a2406f5618973

pallet/xrpl/src/lib.rs
zees-dev: Im not sure if this is correct. Pallet ethereum does not seem to do that.
The validate and apply self contained logic does not overlap.
I am under the assumption that a tx is validated in the mempool - only after validation it is propagated to the nodes to be put in the next block.
The un-validated tx in the mempool is not distributed out.

zees-dev: Ok you are right according to the [docs](https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/primitives/runtime/src/traits.rs#L1497) - whenever we run `pre_dispatch` we also need to run the same validations as `validate`.

zees-dev: addressed: https://github.com/futureversecom/trn-seed/pull/758/commits/8eb136cd65745f892be8a224c2b54dd4f0ac076d

pallet/xrpl/src/types.rs
```diff
@@ -0,0 +1,653 @@
...
+	let mut sha512 = sha2::Sha512::new();
+	sha512.update(message);
+	sha512.finalize()[..SHA512_HASH_LENGTH].try_into().unwrap()
+}
+
+/// The prefix to be prepended to the signed message, for signature verification.
+/// This is hex decoded string of the value `0x53545800`
+pub const SIGNED_MESSAGE_PREFIX: [u8; 4] = [83, 84, 88, 0];
+
+pub fn encode_for_signing(tx_common: &TransactionCommon) -> Result<Vec<u8>, &'static str> {
```
JasonTulp, line 39: nit. This could be implemented on TransactionCommon within xrpl_types

pallet/xrpl/src/types.rs
```diff
@@ -0,0 +1,653 @@
...
+	let mut message_with_prefix =
+		Vec::with_capacity(SIGNED_MESSAGE_PREFIX.len() + serialized_msg.len());
+	message_with_prefix.extend_from_slice(&SIGNED_MESSAGE_PREFIX);
+	message_with_prefix.append(&mut serialized_msg);
+
+	Ok(message_with_prefix)
+}
+
+#[derive(Debug, PartialEq, serde::Deserialize)]
+pub struct Memo {
```
JasonTulp, line 57: nit. Some structs are missing doc comments

pallet/xrpl/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+		outer_call: <T as Config>::RuntimeCall,
+		info: &H160,
+		dispatch_info: &DispatchInfoOf<<T as frame_system::Config>::RuntimeCall>,
+		len: usize,
+	) -> Option<sp_runtime::DispatchResultWithInfo<PostDispatchInfoOf<<T as Config>::RuntimeCall>>> {
+		if let Some(Call::transact { encoded_msg, call, signature }) = outer_call.is_sub_type() {
+			// Pre Dispatch
+			let (nonce, tip) = validate_params::<T>(encoded_msg.as_bytes_ref(), signature.as_bytes_ref(), call)
+				.map_err(|e| {
+					log::error!("⛔️ apply_self_contained: failed to validate params: {:?}", e);
```
JasonTulp, line 214: Should use log! macro which uses LOG_TARGET = "xrpl".
Also some of these error messages could be quite long with two warning emojis :stop_sign: Maybe the inner validation errors could be shortened? Not a biggie though

JasonTulp: eg:
`⛔️ apply_self_contained: failed to validate params: ⛔️ failed to extract extrinsic data from memo data: {memo_data}, err: failed to get chain_id from memo_data`

pallet/xrpl/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+			origin: OriginFor<T>,
+			encoded_msg: BoundedVec<u8, T::MaxMessageLength>,
+			_signature: BoundedVec<u8, T::MaxSignatureLength>,
+			call: Box<<T as Config>::RuntimeCall>,
+		) -> DispatchResult {
+			ensure_none(origin)?;
+
+			let tx: XRPLTransaction = XRPLTransaction::try_from(encoded_msg.as_bytes_ref())
+				.map_err(|e| {
+					log::error!("⛔️ failed to convert encoded_msg to XRPLTransaction: {:?}", e);
```
JasonTulp, line 429: Why log errors within the transaction? Would the pallet error not suffice?

pallet/xrpl/src/lib.rs
surangap: might need to change to `log!()` macro

surangap (approved):
lgtm👍, need to address minors in the comments

ken-centrality:
I wonder if we should use `genesisHash` as a way to identify the chain rather the `chainId` in the signing payload, the problem with `chainId` is it relies entirely on the admin to make sure it's unique, there is nothing on-chain could validate that, meanwhile, genesisHash is a guarantee uniqueness. It also creates a weird coupling dependency between the XRPL and Frontier pallet, I assume.

pallet/xrpl/src/types.rs
zees-dev: could do it - but may be difficult to get this upstreamed.
Specifically the `SIGNED_MESSAGE_PREFIX` var used; without diving deep it's hard to know how the byte array was derived.

pallet/xrpl/src/types.rs
zees-dev: addressed: https://github.com/futureversecom/trn-seed/pull/758/commits/a7a4f0b39d64a7c6b8078a9e587cd5901a115926

pallet/xrpl/src/lib.rs
zees-dev: addressed: https://github.com/futureversecom/trn-seed/pull/758/commits/c06056a065283e041848bd6213bd73dde6f91a47

JasonTulp (approved):
Nice work :rocket:

pallet/xrpl/src/lib.rs
zees-dev: updated to `info` logs

e2e/test/XRPL.test.ts
```diff
@@ -24,20 +24,22 @@ describe("XRPL pallet", () => {
     const wsProvider = new WsProvider(`ws://127.0.0.1:${node.wsPort}`);
     api = await ApiPromise.create({ provider: wsProvider, types: typedefs });
     alith = new Keyring({ type: "ethereum" }).addFromSeed(hexToU8a(ALITH_PRIVATE_KEY));
-    chainId = +(await api.query.evmChainId.chainId());
+    genesisHash = (await api.rpc.chain.getBlockHash(0)).toHex().slice(2);
```
ken-centrality, line 28: just a side note, you can use `api.genesisHash` to get that information

zees-dev: addresed: https://github.com/futureversecom/trn-seed/pull/758/commits/12bcc2a2603b40ea0084e5001b03a9783adce137

PR #759, opened 2023-12-08
Update erc20 precompile transfer to use extrinsic

KarishmaBothara:
**PR checklist**  
Currently some precompiles call the trait internally rather than the extrinsic directly, such as ERC20-transfer. This means that if we block the assets-transfer extrinsic, the precompile will still function normally. If we want to be able to block precompiles with the maintenance-mode block_call extrinsic, these precompiles will need to change their implementation.

We can block precompile addresses from executing which can achieve similar results, however this may not give us the full flexibility we require

I have attached the precompile list which calls trait internally in the ticket

Update ERC20 precompile in this PR to use extrinsic over trait

Closes https://futureverse.atlassian.net/browse/TRN-166


evm-precompiles/erc20/src/lib.rs
```diff
@@ -277,19 +279,26 @@ where
 		handle.record_cost(Runtime::GasWeightMapping::weight_to_gas(
 			<Runtime as pallet_assets::Config>::WeightInfo::transfer(),
 		))?;
-		let _ = <pallet_assets_ext::Pallet<Runtime> as Transfer<Runtime::AccountId>>::transfer(
```
JasonTulp, line 280: We can remove the handle.record_cost above this which manually records the cost of the transfer extrinsic. By using try_dispatch we no longer need this

evm-precompiles/erc20/src/lib.rs
```diff
@@ -339,16 +348,19 @@ where
 			handle.record_cost(Runtime::GasWeightMapping::weight_to_gas(
 				<Runtime as pallet_assets::Config>::WeightInfo::transfer(),
 			))?;
-			let _ = <pallet_assets_ext::Pallet<Runtime> as Transfer<Runtime::AccountId>>::transfer(
```
JasonTulp, line 342: Can remove the manual record_cost above this too

e2e/test/ERC20/Erc20.TxCosts.test.ts
```diff
@@ -216,7 +216,7 @@ describe("ERC20 Gas Estimates", function () {
     const extrinsicFeeCost = balanceBefore.sub(balanceAfter);
     const extrinsicGasScaled = await getScaledGasForExtrinsicFee(provider, extrinsicFeeCost);
 
-    expect(precompileGasEstimate).to.be.lessThan(contractGasEstimate);
+    expect(precompileGasEstimate).to.be.greaterThan(contractGasEstimate);
```
JasonTulp, line 219: Hopefully once the extra gas is removed this will go back to lessThan

JasonTulp (commented):
Looks good! Just need to remove the double gas calculation :)

e2e/test/ERC20/Erc20.TxCosts.test.ts
KarishmaBothara: thanks.. yes, you were right.. it does solves this issue.. :)

evm-precompiles/erc20/src/lib.rs
```diff
@@ -274,22 +276,25 @@ where
...
-		))?;
-		let _ = <pallet_assets_ext::Pallet<Runtime> as Transfer<Runtime::AccountId>>::transfer(
-			asset_id,
-			&origin,
-			&to.clone().into(),
-			amount,
-			false,
-		)
-		.map_err(|e| revert(alloc::format!("ERC20: Dispatched call failed with error: {:?}", e)))?;
+		let destination = to.clone().into();
```
zees-dev, line 279: not sure if we need rhe `.into()` here - as we dont explicitly define the converted type
```suggestion
		let destination = to.clone();
```

e2e/.yarn/install-state.gz
zees-dev, line 1: remove this file? not sure why this is committed in.

zees-dev (changes requested):
LGTM, minor changes required.
Also CI needs to be fixed/passing.

evm-precompiles/erc20/src/lib.rs
KarishmaBothara: this doesn't work without `.into()`
```
--> evm-precompiles/erc20/src/lib.rs:283:18
    |
283 |                 destination: to.clone(),
    |                              ^^^^^^^^^^ expected associated type, found `H160`
    |
    = note: expected associated type `<Runtime as frame_system::Config>::AccountId`
                        found struct `H160`
    = help: consider constraining the associated type `<Runtime as frame_system::Config>::AccountId` to `H160` or calling a method that returns `<Runtime as frame_system::Config>::AccountId`
    = note: for more information, visit https://doc.rust-lang.org/book/ch19-03-advanced-traits.html
help: call `Into::into` on this expression to convert `H160` into `<Runtime as frame_system::Config>::AccountId`
    |
283 |                 destination: to.clone().into(),
    |                                        +++++++
```

zees-dev: Ah interesting, ok ignore the comment then 👍

zees-dev (approved):

PR #761, opened 2023-12-13
Feat/trn 287 xrpl delay

JasonTulp:
Creates delay mechanism for XRPL bridge pallet. If the withdraw amount is over a set limit, it will be delayed by a set delay.


pallet/xrpl-bridge/src/lib.rs
```diff
@@ -783,19 +952,55 @@ impl<T: Config> Pallet<T> {
 			tx_ticket_sequence: ticket_sequence,
 		};
 
-		let proof_id = Self::submit_withdraw_request(door_address.into(), tx_data)?;
+		// Check if there is a payment delay and delay the payment if necessary
+		if let Some((min_balance, delay)) = PaymentDelay::<T>::get() {
+			if amount >= min_balance {
```
surangap, line 957: `min_balance` name seems not suitable for this purpose?  It's the threshold for delaying the payments. in a way it's a max value.

JasonTulp:
@JasonTulp Wrap delay in enum to futureproof if we add deposit delay

pallet/xrpl-bridge/src/lib.rs
```diff
@@ -375,6 +421,27 @@ pub mod pallet {
 			Ok(())
 		}
 
+		/// Sets the payment delay
+		/// payment_delay is a tuple of payment_threshold and delay in blocks
+		#[pallet::weight(0)]
```
surangap, line 426: set the weight

pallet/xrpl-bridge/src/lib.rs
```diff
@@ -633,6 +700,108 @@ impl<T: Config> Pallet<T> {
...
+
+		let mut used_weight = base_process_weight;
+		let highest_processed_delay_block = <NextDelayProcessBlock<T>>::get();
+		let mut new_highest = highest_processed_delay_block;
+		// Limit the number of blocks to process to either the current block_number or the
+		// MaxDelayedPaymentsPerBlock + highest_processed_delay_block
+		let block_limit = block_number.min(
+			highest_processed_delay_block
+				.saturating_add(T::MaxDelayedPaymentsPerBlock::get().into()),
+		);
```
surangap, line 735: Is this correct? `highest_processed_delay_block` is a block number and `T::MaxDelayedPaymentsPerBlock` is number of payments per block.

pallet/xrpl-bridge/src/tests.rs
```diff
@@ -1051,6 +1037,441 @@ fn prune_settled_ledger_index_only_root() {
...
+fn process_delayed_payments_nothing_to_process_works() {
+	TestExt::<Test>::default().build().execute_with(|| {
+		let door = XrplAccountId::from_slice(b"5490B68F2d16B3E87cba");
+		assert_ok!(XRPLBridge::set_door_address(RuntimeOrigin::root(), door));
+		let max_payments_per_block = MaxDelayedPaymentsPerBlock::get() as u64; // 1000
+																	   // Set next process block to 0
+		NextDelayProcessBlock::<Test>::put(0);
+		// Call process delayed payments with enough weight to process 1000 blocks
+		let weight_used = XRPLBridge::process_delayed_payments(
+			max_payments_per_block,
```
surangap, line 1416: does `max_payments_per_block` represent max delayed withdrawals per block? bit confusing to pass it as the block number

pallet/xrpl-bridge/src/tests.rs
```diff
@@ -1051,6 +1037,441 @@ fn prune_settled_ledger_index_only_root() {
...
+fn process_delayed_payments_does_not_exceed_max_delayed_payments() {
+	TestExt::<Test>::default().build().execute_with(|| {
+		let door = XrplAccountId::from_slice(b"5490B68F2d16B3E87cba");
+		assert_ok!(XRPLBridge::set_door_address(RuntimeOrigin::root(), door));
+		let max_payments_per_block = MaxDelayedPaymentsPerBlock::get() as u64; // 1000
+																	   // Set next process block to 0
+		NextDelayProcessBlock::<Test>::put(0);
+		// Call process delayed payments with more than max_payments_per_block
+		let weight_used = XRPLBridge::process_delayed_payments(
+			max_payments_per_block + 10000,
```
surangap, line 1448: same as https://github.com/futureversecom/trn-seed/pull/761/files#r1456593939

surangap (commented):
Implementation looks good. Added a couple of comments to some unclear items

pallet/xrpl-bridge/src/lib.rs
```diff
@@ -70,7 +68,7 @@ pub use weights::WeightInfo;
 #[frame_support::pallet]
 pub mod pallet {
 	use super::*;
-	use seed_primitives::xrpl::XrplTxTicketSequence;
+	use crate::types::DelayedWithdrawal;
```
zees-dev, line 71: nit: redundant import, `super::*` should import this in module

pallet/xrpl-bridge/src/lib.rs
```diff
@@ -809,21 +987,60 @@ impl<T: Config> Pallet<T> {
...
+		sender: T::AccountId,
+		withdrawal: XrpWithdrawTransaction,
+		destination_tag: Option<u32>,
+	) -> DispatchResult {
+		// Get the next payment ID
+		let delayed_payment_id = NextDelayedPaymentId::<T>::get();
+		ensure!(
+			delayed_payment_id.checked_add(One::one()).is_some(),
+			Error::<T>::NoAvailablePaymentIds
+		);
```
zees-dev, line 1016: nit:
```rust
    let next_payment_id = NextDelayedPaymentId::<T>::get()
        .checked_add(One::one())
        .ok_or(Error::<T>::NoAvailablePaymentIds)?;
```
then assign this in the `NextDelayedPaymentId::<T>::put` below?

pallet/xrpl-bridge/src/tests.rs
```diff
@@ -1209,6 +1195,521 @@ fn prune_settled_ledger_index_only_root() {
 	})
 }
 
+#[test]
+fn set_payment_delay_works() {
+	TestExt::<Test>::default().build().execute_with(|| {
```
zees-dev, line 1200: nit: should assert the event emitted in the test

pallet/xrpl-bridge/src/tests.rs
```diff
@@ -1209,6 +1195,521 @@ fn prune_settled_ledger_index_only_root() {
...
+			};
+			// Check balance is reduced
+			let xrp_balance = AssetsExt::balance(XrpAssetId::get(), &account);
+			assert_eq!(xrp_balance, initial_balance - amount);
+
+			// Check storage is correctly mutated
+			assert_eq!(NextDelayedPaymentId::<Test>::get(), delayed_payment_id + 1);
+			assert_eq!(DelayedPayments::<Test>::get(delayed_payment_id), Some(delayed_withdrawal));
+			assert_eq!(
+				DelayedPaymentSchedule::<Test>::get(1001).unwrap().into_inner(),
```
zees-dev, line 1295: nit: would be good to know how `1001` is derived

pallet/xrpl-bridge/src/tests.rs
```diff
@@ -1209,6 +1195,521 @@ fn prune_settled_ledger_index_only_root() {
...
+
+			// Check delay storage is unchanged
+			assert_eq!(NextDelayedPaymentId::<Test>::get(), delayed_payment_id);
+			assert_eq!(DelayedPayments::<Test>::get(delayed_payment_id), None);
+			assert_eq!(DelayedPaymentSchedule::<Test>::get(1001), None);
+		})
+}
+
+#[test]
+fn process_delayed_payments_works() {
```
zees-dev, line 1434: was expecting a test to trigger the on_idle which would process a delayed withdrawal request.
I.e. something like:
- making a withdrawal request
- setting the block number to the blocknumber whereon_idle should successfully process the request
- triggerring the on_idle hook
- validating the storage values and emitted events

pallet/xrpl-bridge/src/lib.rs
```diff
@@ -658,6 +731,111 @@ impl<T: Config> Pallet<T> {
...
+		let mut new_highest = highest_processed_delay_block;
+		// Limit the number of blocks to process to either the current block_number or the
+		// DelayedPaymentBlockLimit + highest_processed_delay_block
+		let block_limit = block_number
+			.min(highest_processed_delay_block.saturating_add(T::DelayedPaymentBlockLimit::get()));
+
+		// Get the current door address
+		let Some(door_address) = DoorAddress::<T>::get() else {
+			return used_weight;
+		};
```
zees-dev, line 769: nit:
```rust
let door_address = DoorAddress::<T>::get().ok_or(used_weight)?;
```

zees-dev (approved):
Great work here!

A few small nits, maybe some comments to take a quick glance at.
Otherwise LGTM! 💯

pallet/xrpl-bridge/src/lib.rs
JasonTulp: I agree this is cleaner, although we use the flow above in many other places where the stored value represents the Next available ID, not the highest used ID

pallet/xrpl-bridge/src/lib.rs
JasonTulp: This only works when returning a Result

pallet/xrpl-bridge/src/tests.rs
JasonTulp: The reason I bypassed on_idle for tests is so I could more accurately record and check the weight being returned, the on_idle function in xrpl-bridge is overworked and handles the pruning of tx data as well so the on_idle will return weight for checking that (Even if there is nothing to return) That being said, I've added a test that calls on_idle without verifying exact weight used

surangap (approved):
lgtm

PR #763, opened 2023-12-29
fix: Futurepass precompile custom revert message prefix

zees-dev:
## Changelog

All precompile custom revert messages should be prefixed with the pre-compile name.
This is especially helpful when debugging across multiple contracts (as a revert in a call may be too ambiguous to figure out where the transaction is failing).

- Updated `Futurepass` pre-compile revert messages to be prefixed with `"Futurepass: "`
- Updated `FuturepassRegistrar` pre-compile revert messages to be prefixed with `"Futurepass Registrar: "`
- Fixed bug in `FuturepassRegistrar` pre-compile `create(address)` function
  - function signature is: `function create(address owner) external returns (address);`, but it does not return back the created futurepass address


JasonTulp (approved):
Nice change

PR #764, opened 2024-01-08
Update the bootnodes list for both Root and Porcini

ken-centrality:
Update the bootnodes list for Root and Porcini based on the list here

https://github.com/futureversecom/aws-infra-cdk/blob/features/init-cdk/thousand-sunny/src/bin/run-node

**PR checklist**  


juehai:
The Bootnode list looks good to me. BTW, do we want to update the `root_human.json` with the same list as well ?

juehai (approved):

ken-centrality:
Managed to connect up to 9 nodes with this image.

PR #765, opened 2024-01-10
Remove create fee override

JasonTulp:
Removes the 2,000,000 evm gas_transaction_create override in favor of the default value. This will bring the cost of creating contracts in line with other EVM compatible chains


zees-dev:
Should probably add an e2e test for this. 

It maybe easier to modify an existing e2e test and assert the lowered cost of contract creation/deployment.

surangap (approved):
lgtm, never saw that override 😁

zees-dev (approved):
Lgtm

PR #767, opened 2024-01-15
Fix Asset Existence Verification in ERC20 Precompile

JasonTulp:
Currently the ERC20 precompile checks if an address is a precompile address by checking whether the asset exists. It does this by checking the total supply of an asset which is not always accurate as an asset can be created with 0 supply.

This has been changed to check the owner field of an asset, all assets will have an account stored in owner so this is far more accurate


zees-dev (commented):
Changes look good.

I would advise having atleast a unit test for this.

Something to note however.
Currently our ERC20 does not support `Ownable` - but one day we may have to.
In such a case ownership could be transferred to a potential dead address (0 address).
This may imply no owner - not sure if we need to handle this case however (food for thought).

zees-dev (approved):

PR #768, opened 2024-01-16
Feat/290 doughnut self contained tests

JasonTulp:
Adds signed validation tests to doughnut PR. 
Refactors accounts for tests to simplify code


surangap (approved):
good tests. lgtm

PR #772, opened 2024-01-23
Doughnut benchmarks

JasonTulp:
Adds benchmarks to doughnut pallet for all 3 extrinsics. 
Note, benchmarks use hardcoded encoded doughnut due to constraints with sp_core imports.


surangap (approved):
lgtm, might need a final thought on how we write benches overall.

PR #776, opened 2024-02-04
Change origin for doughnut.transact

JasonTulp:
Changes the origin check to ensure_none for doughnut.transact. This prevents other extrinsics proxying doughnut transact as the inner call


pallet/doughnut/src/lib.rs
```diff
@@ -352,7 +351,6 @@ pub mod pallet {
 			// Check holder == sender
 			let issuer_address = Self::get_address(doughnut_v1.issuer)?;
 			let holder_address = Self::get_address(doughnut_v1.holder)?;
-			ensure!(holder_address == sender, Error::<T>::UnauthorizedSender);
```
surangap, line 355: `sender == holder` check might be required somewhere in the flow since we only allow the holder to call with doughnuts. But this work will be touched with `holder whitelisting work`. Can be addressed then.

surangap (approved):
lgtm

pallet/doughnut/src/lib.rs
JasonTulp: I believe this is covered by the outer signature verification. The sender == holder check here was actually pointless as we were just passing through the holder_address anyway

PR #783, opened 2024-02-14
Feat/sft marketplace support

JasonTulp:
- Adds support for buying and selling SFTs in the marketplace
- Adds marketplace_id to listing events
- Changes event token type from collection_id + serial_numbers to listing_tokens
- Changes Listing struct to include listing_tokens (Supports both SFT and NFT)


pallet/marketplace/src/types.rs
```diff
@@ -34,6 +38,137 @@ pub type OfferId = u64;
...
+#[scale_info(skip_type_params(T))]
+pub struct SftListing<T: Config> {
+	pub collection_id: CollectionUuid,
+	pub serial_numbers: BoundedVec<(SerialNumber, Balance), <T as Config>::MaxTokensPerListing>,
+}
+
+impl<T: Config> ListingTokens<T> {
+	/// Validates the listing tokens by checking the following:
+	/// - Ensures the list of tokens is not empty
+	/// - Ensures all tokens within the listing are from the same collection
```
surangap, line 68: should this be?
```suggestion
	/// - Ensures all tokens within the listing are from the same type. i.e nft or sft
```

JasonTulp: Actually this line should be removed entirely. This function previously validated the collectionId was the same across a list of token_ids but that is now enforced with a changed data structure. Will remove :)

runtime/src/migrations/marketplace.rs
zees-dev, line 1: please fork-test this migration before release

pallet/sft/src/traits.rs
zees-dev, line 1: The trait should not be in this pallet - since if another pallet needs to use the SFT functionality; it will need to pull in the whole pallet just to pull in this trait.
If it's already uses the pallet as a dependency, then there isn't a need to use the trait - since there is already tight-coupling introduced.

This needs to go in a seperate pallet.

runtime/src/migrations/marketplace.rs
JasonTulp: Have tested yesterday

pallet/marketplace/src/impls.rs
```diff
@@ -370,7 +355,11 @@ impl<T: Config> Pallet<T> {
 		ensure!(offer.buyer == who, Error::<T>::NotBuyer);
 		T::MultiCurrency::release_hold(T::PalletId::get(), &who, offer.asset_id, offer.amount)?;
 		let _ = Self::remove_offer(offer_id, offer.token_id)?;
-		Self::deposit_event(Event::<T>::OfferCancel { offer_id, token_id: offer.token_id });
+		Self::deposit_event(Event::<T>::OfferCancel {
```
zees-dev, line 358: how do we know which collection the token is from?

pallet/sft/src/traits.rs
JasonTulp: Agreed: [f2316996498cefc485ad8cc356d5c591ddffb548](https://github.com/futureversecom/trn-seed/pull/783/commits/f2316996498cefc485ad8cc356d5c591ddffb548)

pallet/marketplace/src/impls.rs
```diff
@@ -393,24 +382,25 @@ impl<T: Config> Pallet<T> {
...
 			offer.asset_id,
-			collection_id,
-			serial_numbers,
+			listing_tokens,
 			offer.amount,
 			royalties_schedule,
 		)?;
 
 		let _ = Self::remove_offer(offer_id, offer.token_id)?;
 		Self::deposit_event(Event::<T>::OfferAccept {
```
zees-dev, line 401: Same here, don't we need the collection id too

pallet/marketplace/src/impls.rs
JasonTulp: TokenId is made up of (CollectionId, SerialNumber)

pallet/marketplace/src/impls.rs
JasonTulp: collection_id is contained within listing_tokens

pallet/marketplace/src/tests.rs
```diff
@@ -2402,3 +2497,1406 @@ mod set_fee_to {
...
+					None,
+					None,
+				),
+				Error::<Test>::RoyaltiesInvalid
+			);
+		});
+	}
+
+	#[test]
+	fn sell_sft_duplicate_serial_numbers() {
```
zees-dev, line 2720: nit: could also validate event

pallet/marketplace/src/tests.rs
zees-dev, line 1: nit: some of the newer tests could do with event validations.
Or that could simply move to benchmarking tests - since you are testing every single extrinsics success.

pallet/sft/src/traits.rs
zees-dev: Note: The same should be done for the NFTExt too; not sure if in this PR or another

zees-dev (approved):
LGTM; nice work!

pallet/marketplace/src/tests.rs
JasonTulp: I only validate events once in most cases as I don't see the need to validate in every test (As it reduces readability of tests and imo is unecessary in a lot of cases)

JasonTulp: Happy to discuss best approach for this though

pallet/marketplace/src/tests.rs
JasonTulp: Addressed: [265a791d46929cdf3b35475004f054ceb5b68c8b](https://github.com/futureversecom/trn-seed/pull/783/commits/265a791d46929cdf3b35475004f054ceb5b68c8b)

pallet/sft/src/traits.rs
JasonTulp: I had a quick go at migrating NFTExt but it's not as simple because it's a bit more widespread. Will address this in a separate PR for times sake

pallet/marketplace/src/tests.rs
zees-dev: Once is fine; id agree 👍

surangap (approved):
lgtm👍

PR #785, opened 2024-02-15
Add logs

surangap:
### Changelog
- Add some error logs in the self-contained path.

### DIFF
--- a/pallet/doughnut/src/lib.rs
+++ b/pallet/doughnut/src/lib.rs
@@ -33,12 +33,11 @@ use doughnut_rs::{
 };
 use frame_support::{
 	dispatch::{DispatchInfo, GetDispatchInfo, PostDispatchInfo},
-	log,
 	traits::{GetCallMetadata, IsSubType},
 };
 use frame_system::{CheckNonZeroSender, CheckNonce, CheckWeight};
 use pallet_transaction_payment::{ChargeTransactionPayment, OnChargeTransaction};
-use seed_pallet_common::ExtrinsicChecker;
+use seed_pallet_common::{log, ExtrinsicChecker};
 use seed_primitives::AccountId20;
 use sp_runtime::{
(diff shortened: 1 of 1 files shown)


pallet/doughnut/src/lib.rs
```diff
@@ -81,10 +85,14 @@ impl<T> Call<T>
 					return Err(TransactionValidityError::Invalid(InvalidTransaction::BadProof))
 				};
 				// Verify doughnut signature
-				doughnut_v1.verify().map_err(|_| TransactionValidityError::Invalid(InvalidTransaction::BadProof))?;
+				doughnut_v1.verify().map_err(|e| {
+					log::error!("⛔️ failed to verify doughnut signature: {:?}", e);
```
surangap, line 89: thinking if we should make it debug only. somebody could play with the nodes for free

pallet/doughnut/src/lib.rs
```diff
@@ -137,12 +148,15 @@ impl<T> Call<T>
...
+				log::error!("⛔️ failed to get fee payer address: {:?}", doughnut_v1.fee_payer());
 				return None
 			};
 			let mut fee_payer_address = fee_payer_doughnut;
+
 			// Futurepass check
 			if <T as Config>::FuturepassLookup::check_extrinsic(inner_call, &()).is_ok() {
-				let Ok(futurepass) = <T as Config>::FuturepassLookup::lookup(fee_payer_address.into()) else {
+				let Ok(futurepass) = <T as Config>::FuturepassLookup::lookup(fee_payer_address.clone().into()) else {
+					log::error!("⛔️ failed to retrieve futurepass address for the address: {:?}", fee_payer_address);
```
JasonTulp, line 159: Should we add `target: "doughnut"` or is that covered by the LOG_TARGET constant?

surangap: yeah wrong macro. updated. thanks

PR #790, opened 2024-02-22
Pallet Crowdsale

zees-dev:
# Pallet Crowdsale pallet

An on-chain, in-protocol crowdsale pallet which allows permissionless creation of crowdsales via an account.
Users can participate in these open crowdsales provided that they have they required payment asset.
The crowdsale rewards (vouchers/assets) will be automatically distributed (also manually claimable) after the conclusion of the crowdsales.
The vouchers are fungible assets (can be traded, deposited on DEX, etc.) that can then be redeemed for NFT's promised by the crowdsale account.

### `initialize`
- Permissionless setup a crowdsale by any account
- An NFT `collection_id` is passed in here which must be created prior to this call
- NFT collection `max_issuance` must be set (This is the final max issuance of the crowdsale) - it works in tandem with the `soft_cap_price` to determine crowdsale target
- NFT `collection_issuance` must be `0` (i.e. No tokens already minted)
- Initialize will transfer ownership of the NFT collection to an auto-generated Vault account so metadata MUST be correct before calling this step
- A voucher Asset and ID will be automatically generated at this step
- The entire total supply of vouchers will also be minted to the vault account
- The duration must be below 3 months (90 days)
### `enable`
- Enables and starts the crowdsale, anybody will be able to participate in the sale past this point
- Callable only by the sale admin account (initializer of the sale)
- Once enabled, the crowdsale will automatically end after the specified duration
### `participate`
- Pay funds to the vault account to participate in the crowdsale
- Funds are not refundable but can be accumulated across multiple transactions
- Vouchers claimable at the end of the sale is determined by amount participated
### `claim_voucher`
- Manually claim a voucher after a crowdsale ends
- Even though vouchers are automatically distributed, participants can bypass this and manaully call `claim_voucher` to receive their vouchers instantly at the cost of gas
### `redeem_voucher`
- Exchange vouchers for NFTs
- NFTs are minted to the caller's account upon successful redemption
- `quantity` is quantity of NFTs redeemed, NOT vouchers (i.e. without decimal places)
- Voucher tokens are burnt at this stage
### `try_force_distribution`
- In the unlikely event that network traffic is so high that automatic distribution trigger fails, this can be called to retry auto-distribution
### `distribute_crowdsale_rewards`
- Called from the `offchain_worker` to automatically distribute vouchers after a sale ends
- Distributes 100 vouchers every ~8 blocks
- First in First out (I.e. if multiple payouts occur simultaneously, they will be paid out sequentially)     

## TODO
- scripts (load-testing)


JasonTulp:
check coverage

pallet/crowdsale/src/lib.rs
```diff
@@ -0,0 +1,773 @@
+// Copyright 2023-2024 Futureverse Corporation Limited
```
surangap, line 1: nit- should the copyright be 2024 onwards since the code is added in 2024?

pallet/crowdsale/src/lib.rs
```diff
@@ -0,0 +1,773 @@
...
+		/// Crowdsale NFT redeemed
+		CrowdsaleNFTRedeemed {
+			sale_id: SaleId,
+			who: T::AccountId,
+			collection_id: CollectionUuid,
+			quantity: TokenCount,
+		},
+		/// Crowdsale closed
+		CrowdsaleClosed { sale_id: SaleId, info: SaleInformation<T::AccountId, T::BlockNumber> },
+		/// Crowdsale distribution was manually re triggered
```
surangap, line 180: /// Crowdsale distribution was manually triggered

surangap, line 180:
```suggestion
		/// Crowdsale distribution was manually triggered
```

pallet/crowdsale/src/impls.rs
```diff
@@ -0,0 +1,288 @@
...
+						.ok_or(Error::<T>::MaxIssuanceNotSet)?;
+				let collection_max_issuance: Balance = collection_max_issuance.into();
+
+				// Should find the voucher price
+				let crowd_sale_target =
+					sale_info.soft_cap_price.saturating_mul(collection_max_issuance);
+
+				if sale_info.funds_raised < crowd_sale_target {
+					// Refunded amount is equal to the total issuance minus the total vouchers paid
+					// out Total vouchers paid out is the total funds raised divided by the voucher
```
surangap, line 244:
```suggestion
					// out. Total vouchers paid out is the total funds raised divided by the voucher
```

pallet/crowdsale/src/impls.rs
```diff
@@ -0,0 +1,288 @@
...
+		Self::deposit_event(Event::CrowdsaleVouchersClaimed {
+			sale_id,
+			who,
+			amount: claimable_vouchers,
+		});
+		Ok(claimable_vouchers)
+	}
+
+	/// Close all crowdsales that are scheduled to end this block.
+	pub(crate) fn close_sales_at(now: T::BlockNumber) -> Result<u32, &'static str> {
```
surangap, line 206: Let's return the weight in this function. would be convenient to calculate it here.

pallet/crowdsale/src/impls.rs
```diff
@@ -0,0 +1,288 @@
...
+// You may obtain a copy of the License at the root of this project source code
+
+use crate::*;
+use alloc::format;
+use frame_support::sp_runtime::traits::{BlakeTwo256, Hash};
+use sp_core::U256;
+
+impl<T: Config> Pallet<T> {
+	/// Create a unique account (vault) to hold funds (payment_asset) for a crowdsale.
+	pub(crate) fn vault_account(nonce: SaleId) -> T::AccountId {
```
surangap, line 19: Maybe can modify the function in a way that can be moved to a pallet-common ?

pallet/crowdsale/src/lib.rs
```diff
@@ -0,0 +1,773 @@
...
+				T::MultiCurrency::transfer(
+					sale_info.payment_asset_id,
+					&who,
+					&sale_info.vault,
+					amount,
+					false,
+				)?;
+
+				// update the sale funds
+				sale_info.funds_raised = sale_info.funds_raised.saturating_add(amount);
```
surangap, line 470: Do you think we can remove this and hence mutating `SaleInfo` inside `participate()`?
instead let's do a read here and update the `SaleInfo` at the sale close time? can use `SaleParticipation` for this.

pallet/crowdsale/src/impls.rs
zees-dev: Maybe if we need again; will require additional thought since `T::PalletId` is not always a type on config trait.
We may need 2 params for the function (nonce + some seed).
I cannot seem to make it generic enough to allow re-use of it in the common pallet.

pallet/crowdsale/src/lib.rs
zees-dev: https://github.com/futureversecom/trn-seed/pull/790/commits/651e3f3ef6c27c71d7086a03fab57b58864cfb18

pallet/crowdsale/src/lib.rs
zees-dev: https://github.com/futureversecom/trn-seed/pull/790/commits/c81da1c363159df092b64c9e66af2d90cdbdc397

pallet/crowdsale/src/impls.rs
zees-dev: https://github.com/futureversecom/trn-seed/pull/790/commits/c81da1c363159df092b64c9e66af2d90cdbdc397

pallet/crowdsale/src/lib.rs
```diff
@@ -0,0 +1,775 @@
...
+		#[transactional]
+		pub fn redeem_voucher(
+			origin: OriginFor<T>,
+			sale_id: SaleId,
+			quantity: TokenCount,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			ensure!(quantity > 0, Error::<T>::InvalidQuantity);
+
+			SaleInfo::<T>::try_mutate(sale_id, |sale_info| -> DispatchResult {
```
surangap, line 692: do we need to mutate here? simple read should be sufficient?

zees-dev: good catch!
addressed: https://github.com/futureversecom/trn-seed/pull/790/commits/f3f109c2c4c04b87b80f20cede1417da88156845

leo-plugdefi (approved):

surangap (approved):
lgtm. nice work 🚀

PR #724, opened 2023-10-26
Migrate token ownership to separate storage map

JasonTulp:
This PR moves token ownership in the NFT pallet to it's own map. This allows for easier querying of the collection_info struct and avoids massive delays when querying large sets of data. The owned_tokens RPC call should be used to get ownership information for individual accounts.

Jira link: [TRN-240](https://futureverse.atlassian.net/jira/software/c/projects/TRN/boards/39?selectedIssue=TRN-240)
Note, also fixes [TRN-225](https://futureverse.atlassian.net/jira/software/c/projects/TRN/boards/39?selectedIssue=TRN-225)

[TRN-240]: https://futureverse.atlassian.net/browse/TRN-240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TRN-225]: https://futureverse.atlassian.net/browse/TRN-225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #792, opened 2024-02-22
Add collection info - RPC call

KarishmaBothara:
Description

As the NFT collection grow in size with new token minted in it.. Storage to get collection info fails to get the huge data. This PR provides RPC method to get collection information from collection id. It has all the information except for collection owner

**PR checklist**  

- [ ] Tag related issue(s)
- [ ] Add appropriate unit tests and/or integration tests to the [integration test repository](https://github.com/futureversecom/seed-integration-tests)


pallet/nft/src/impls.rs
```diff
@@ -361,6 +362,61 @@ impl<T: Config> Pallet<T> {
 		(new_cursor, total_owned, response)
 	}
 
+	/// Find the tokens owned by an `address` in the given collection
+	/// limit return tokens that are larger than the cursor
+	/// Returns list of tokens, the sum of all tokens owned by the user
+	/// and the new cursor for the next owned SerialNumber not included in the returned list
```
JasonTulp, line 368: Comment needs updating

pallet/nft/src/impls.rs
```diff
@@ -361,6 +362,61 @@ impl<T: Config> Pallet<T> {
...
+	pub fn collection_info(
```
JasonTulp, line 369: The name of this function is a little ambiguous and may be confused with the collection_info storage map

pallet/nft/src/impls.rs
```diff
@@ -361,6 +362,61 @@ impl<T: Config> Pallet<T> {
...
+			Zero::zero()
+		};
+
+		let max_issuance = collection_info.max_issuance;
+		let next_serial_number = collection_info.next_serial_number;
+		let collection_issuance = collection_info.collection_issuance;
+		let cross_chain_compatibility = collection_info.cross_chain_compatibility.xrpl;
+
+		(
+			T::AccountId::from(owner),
```
JasonTulp, line 409: I don't think this from is needed as owner should already be T::AccountId?

pallet/nft/src/impls.rs
```diff
@@ -361,6 +362,61 @@ impl<T: Config> Pallet<T> {
...
+		let royalties_schedule = if collection_info.royalties_schedule.is_some() {
+			collection_info.royalties_schedule.unwrap().calculate_total_entitlement()
+		} else {
+			Zero::zero()
+		};
+
+		let max_issuance = collection_info.max_issuance;
+		let next_serial_number = collection_info.next_serial_number;
+		let collection_issuance = collection_info.collection_issuance;
+		let cross_chain_compatibility = collection_info.cross_chain_compatibility.xrpl;
```
JasonTulp, line 406: This field should be the enum variant if we want to expand upon this enum in future. i.e. it doesn't make much sense returning a bool for cross_chain_compatibility (Being the XRPL compatibility) Either we:
- Return the whole enum
- Leave this parameter out entirely (Maybe the best, might need further discussion)
- Rename to XRPL-Compatible

pallet/nft/src/impls.rs
```diff
@@ -361,6 +362,61 @@ impl<T: Config> Pallet<T> {
...
+		let name = collection_info.name.into();
+		let metadata_scheme = collection_info.metadata_scheme.construct_token_uri(0);
+		let royalties_schedule = if collection_info.royalties_schedule.is_some() {
+			collection_info.royalties_schedule.unwrap().calculate_total_entitlement()
+		} else {
+			Zero::zero()
+		};
+
+		let max_issuance = collection_info.max_issuance;
+		let next_serial_number = collection_info.next_serial_number;
```
JasonTulp, line 404: This parameter is mainly for internal use, not that necessary for a collection owner to know. It is used as a counter for assigning next serial numbers. CollectionIssuance should be enough

pallet/nft/src/impls.rs
```diff
@@ -361,6 +362,61 @@ impl<T: Config> Pallet<T> {
...
+				Default::default(),
+			)
+		}
+
+		let collection_info = collection_info.unwrap();
+		let owner = collection_info.owner;
+		let name = collection_info.name.into();
+		let metadata_scheme = collection_info.metadata_scheme.construct_token_uri(0);
+		let royalties_schedule = if collection_info.royalties_schedule.is_some() {
+			collection_info.royalties_schedule.unwrap().calculate_total_entitlement()
```
JasonTulp, line 398: I don't think passing the total entitlement is that useful here. A collection owner is more likely to need the accounts and their individual entitlements than the total entitlements. Maybe this is a good use case for separating out each parameter into it's own RPC call? i.e. get_royalties_schedule

pallet/nft/src/impls.rs
```diff
@@ -361,6 +362,61 @@ impl<T: Config> Pallet<T> {
...
+				Default::default(),
+				Default::default(),
+				Default::default(),
+			)
+		}
+
+		let collection_info = collection_info.unwrap();
+		let owner = collection_info.owner;
+		let name = collection_info.name.into();
+		let metadata_scheme = collection_info.metadata_scheme.construct_token_uri(0);
```
JasonTulp, line 396: We probably don't need this field. Returning the token_uri for token 0 can be achieved using the existing RPC method (get_token_uri) This feels a little redundant

JasonTulp (commented):
Looking good, some fields I don't think are needed in this RPC call. I also think we should have a think about whether it would make sense to have one RPC for all information, or multiple RPC calls for each individual part. What do you think?

KarishmaBothara:
> Looking good, some fields I don't think are needed in this RPC call. I also think we should have a think about whether it would make sense to have one RPC for all information, or multiple RPC calls for each individual part. What do you think?

Thanks for your review.. I think the reason for this rpc call is, storage collection_info would fail, ownedTokens list if it is too huge, then the storage retrieval fails.. so an alternative to that without ownedTokens info

pallet/nft/src/impls.rs
ken-centrality: it should return like this

```
  crossChainCompatibility: {
    xrpl: false
  }
```

pallet/nft/src/impls.rs
ken-centrality: it's might be different when query the bridged collection @JasonTulp ? check out the collection `1124`

```
  nextSerialNumber: 0
  collectionIssuance: 1,214
```

pallet/nft/src/impls.rs
ken-centrality: can we show it like this?

```
  royaltiesSchedule: {
    entitlements: [
      [
        0x8d916dc7279C6646743B7a5EDC3618c3b06DF329
        2.50%
      ]
    ]
  }
```

pallet/nft/src/impls.rs
ken-centrality: I think it's useful @JasonTulp, at the end of the day, user might not know the connection betweem `metadata_scheme` and `token_uri`, as they have quite different naming

pallet/nft/src/impls.rs
KarishmaBothara: done

pallet/nft/src/impls.rs
KarishmaBothara: I have lot of problem here for Serialization and Deserialization...
I tried adding similar to crossChainCompatibility and originChain. But it 
```

// Describes the royalty scheme for secondary sales for an NFT collection/token
#[derive(Debug, Clone, Encode, Decode, PartialEq, Eq, TypeInfo, MaxEncodedLen)]
pub struct RoyaltiesSchedule<AccountId> {
	/// Entitlements on all secondary sales, (beneficiary, % of sale price)
	pub entitlements: BoundedVec<(AccountId, Permill), ConstU32<MAX_ENTITLEMENTS>>,
}
```
internally has boundedVec so have serialize/deserialize BoundedVec as well which is a problem

pallet/nft/src/types.rs
```diff
@@ -75,12 +76,59 @@ where
...
+// impl<'a> Deserialize<'a> for CrossChainCompatibility {
+// 	fn deserialize<D>(deserializer: D) -> Result<CrossChainCompatibility, D::Error>
+// 		where
+// 			D: Deserializer<'a>,
+// 	{
+// 		deserializer.deserialize_any(CrossChainCompatibilityVisitor)
+// 	}
+// }
+//
+// struct CrossChainCompatibilityVisitor;
```
JasonTulp, line 105: nit. Leftover comments

e2e/test/NftRPC.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    const chainOrigin = result[8];
+    expect(owner).to.eql("0xf24ff3a9cf04c71dbc94d0b566f7a27b94566cac");
+    expect(name).to.eql("test-collection");
+    expect(metadata).to.eql("https://test/api/0");
+    expect(royalitySchedule.toJSON()).to.eql([[alith.address, 10000]]);
+    expect(maxIssuance.toJSON()).to.eql(null);
+    expect(nextSerialNumber.toNumber()).to.eql(0);
+    expect(collectionIssuance.toNumber()).to.eql(0);
+    expect(crossChainCompatibility.toJSON()).to.eql({ xrpl: false });
+    expect(chainOrigin.toString()).to.eql("Root");
```
JasonTulp, line 119: How easy do you think it would be to return a struct with custom deserialize that will return a json object with each parameter? Then instead of just a tuple it would look like:
```
{
  owner: "0xf24ff3a9cf04c71dbc94d0b566f7a27b94566cac",
  name: "test-collection",
  metadata: "https://test/api/0",
  ...
}
```
Not a biggie but may be a nicer user experience then pulling out each value based on index

JasonTulp: One way that maaaay work is implementing serialize for CollectionInfo, then skipping the owned_tokens part. 
That or we merge [this PR](https://github.com/futureversecom/trn-seed/pull/724) first
Or we could create a separate struct that encapsulates all the relevant data (But may be unnecessary duplicate code)

pallet/nft/src/impls.rs
```diff
@@ -376,6 +377,58 @@ impl<T: Config> Pallet<T> {
...
+	/// Find the tokens details for the given collection id
+	/// Returns collection owner, name, metadata schema, max issuance,
+	/// next available serial number, collection issuance, is_cross_chain_compatible
+	pub fn collection_details(collection_id: CollectionUuid) -> CollectionDetail<T::AccountId>
+	where
+		<T as frame_system::Config>::AccountId: core::default::Default,
+	{
+		let collection_info = <CollectionInfo<T>>::get(collection_id);
+		if collection_info.is_none() {
+			// should not happen
```
JasonTulp, line 389: There is a chance that an invalid collection_id was supplied to the RPC so this could be None, we should handle this situation and return an error/ None instead of Default

pallet/nft/src/impls.rs
```diff
@@ -376,6 +377,58 @@ impl<T: Config> Pallet<T> {
...
+			let collection_info = CollectionDetail {
+				owner: T::AccountId::from(Default::default()),
+				royalties_schedule: Default::default(),
+				max_issuance: None,
+				origin_chain: Default::default(),
+				next_serial_number: 0,
+				collection_issuance: 0,
+				cross_chain_compatibility: Default::default(),
+				name: Default::default(),
+				metadata_scheme: Default::default(), // MetadataScheme::try_from(b"".as_slice()).unwrap()
```
JasonTulp, line 400: nit, comment left in

pallet/nft/src/types.rs
```diff
@@ -75,18 +76,65 @@ where
...
+	}
+}
+
 impl Default for CrossChainCompatibility {
 	fn default() -> Self {
 		Self { xrpl: false }
 	}
 }
 
+/// Information related to a specific collection
```
JasonTulp, line 102: Could you please add more details in the comment explaining why we needed a separate struct for the RPC result?

runtime/src/lib.rs
```diff
@@ -1472,6 +1472,7 @@ pub type Executive = frame_executive::Executive<
 	AllPalletsWithSystem,
 	(migrations::AllMigrations,),
 >;
+pub const COLLECTION_NAME_STRING_LIMIT: u32 = 50;
```
JasonTulp, line 1475: What is this const used for?

pallet/nft/src/impls.rs
```diff
@@ -376,6 +377,58 @@ impl<T: Config> Pallet<T> {
...
+			};
+			return collection_info;
+		}
+		let collection_info = collection_info.unwrap();
+		let owner = collection_info.owner;
+		let name = collection_info.name.into();
+		let metadata_scheme = collection_info.metadata_scheme.construct_token_uri(0);
```
JasonTulp, line 407: nit, maybe instead of appending 0 we could return the raw token uri? 
```suggestion
		let metadata_scheme = collection_info.metadata_scheme.0.into_inner();
```

pallet/nft/rpc/runtime-api/Cargo.toml
```diff
@@ -9,6 +9,7 @@ repository.workspace = true
 [dependencies]
 codec = { workspace = true }
 sp-api = { workspace = true }
+sp-core = { workspace = true }
```
JasonTulp, line 12: sp-core isn't being used in this crate so can be removed. Also sp-runtime can be removed by the looks of it

JasonTulp (commented):
Looking good! The separate struct looks like the better option for sure. Few minor comments left to address

pallet/nft/rpc/runtime-api/Cargo.toml
```diff
@@ -19,6 +19,7 @@ default = ["std"]
 std = [
 	"codec/std",
 	"sp-api/std",
+	"codec/std",
```
JasonTulp, line 22: Can sp-ruintime be removed from the cargo.toml?

pallet/nft/src/impls.rs
```diff
@@ -376,6 +377,45 @@ impl<T: Config> Pallet<T> {
...
+	/// next available serial number, collection issuance, is_cross_chain_compatible
+	pub fn collection_details(
+		collection_id: CollectionUuid,
+	) -> Result<CollectionDetail<T::AccountId>, DispatchError>
+	where
+		<T as frame_system::Config>::AccountId: core::default::Default,
+	{
+		let mut collection_info =
+			<CollectionInfo<T>>::get(collection_id).ok_or(Error::<T>::NoCollectionFound)?;
+		let collection_info = collection_info;
```
JasonTulp, line 391:
```suggestion
		let collection_info =
			<CollectionInfo<T>>::get(collection_id).ok_or(Error::<T>::NoCollectionFound)?;
```

JasonTulp (approved):
lgtm, 2 more small comments. Needs a review from someone else before merging. Nice work Karishma

pallet/nft/src/impls.rs
```diff
@@ -376,6 +377,45 @@ impl<T: Config> Pallet<T> {
...
+	/// Find the tokens details for the given collection id
+	/// Returns collection owner, name, metadata schema, max issuance,
+	/// next available serial number, collection issuance, is_cross_chain_compatible
+	pub fn collection_details(
+		collection_id: CollectionUuid,
+	) -> Result<CollectionDetail<T::AccountId>, DispatchError>
+	where
+		<T as frame_system::Config>::AccountId: core::default::Default,
+	{
+		let collection_info =
```
surangap, line 389: Do you think this would have the same strain on the node as querying via the chain state? What was the error when the state query failed? I guess if it's the response payload being maxed out, then the rpc would provide a fix for that. better to do a fork test on mainnet against largest collections.

surangap (commented):
RPC changes lgtm. Added a comment -> https://github.com/futureversecom/trn-seed/pull/792/files#r1908075147

pallet/nft/src/impls.rs
KarishmaBothara: Some collection are too big for js to decode it. so, when querying storage ends up with <unknown>
Screenshot 2025-01-09 at 10 20 14 AM
If the encoded key at this storage is used - `0xf43ffbe61ef468749d3617ac1a63c4b7bacdccfb1aa10c1f6eaca6b1835c2036b8eb52f4279dfc96641c0000`
and rpc call used to get data at this storage, i.e
```
curl https://root.au.rootnet.live/ -X POST -H "Content-Type: application/json" --data '{"method":"state_getStorage","params":["0xf43ffbe61ef468749d3617ac1a63c4b7bacdccfb1aa10c1f6eaca6b1835c2036b8eb52f4279dfc96641c0000"],"id":1,"jsonrpc":"2.0"}'
```
This gives encoded output which is way too big, to decode.
RPC solves the issue, as it gives all the info of the collection except ownedToken which takes most storage.
OwnedToken can be fetched using separate rpc call - nft_ownedTokens.

KarishmaBothara: Exact error when querying storage is

```
Error: Unable to decode storage nft.collectionInfo: entry 0:: createType(PalletNftCollectionInformation):: decodeU8aStruct: failed at 0x04b2e85a2e0fb0c8e02d0fddca19ac54… on ownedTokens (index 10/10): Vec<PalletNftTokenOwnership>:: decodeU8aVec: failed at 0xb2e85a2e0fb0c8e02d0fddca19ac54db… (index 1/1): {"owner":"SeedPrimitivesSignatureAccountId20","ownedSerials":"Vec<u32>"}:: decodeU8aStruct: failed at 0x82a83100000000000100000002000000… on ownedSerials (index 2/2): Vec<u32>:: Vec length 813600 exceeds 65536
    at S._newType (https://portal.rootnet.live/modu.26932a01d2e4445d.js:1:204648)
    at S._formatStorageSetEntry (https://portal.rootnet.live/modu.26932a01d2e4445d.js:1:204146)
    at https://portal.rootnet.live/modu.26932a01d2e4445d.js:1:203900
    at a.reduce (<anonymous>)
```

surangap: looks like the decode error here is at the js side. let's see a fork test on mainnet as well.

PR #793, opened 2024-02-22
Pallet crowdsale upd

zees-dev:
# Improvements
- added log_target
- decimals for voucher is based on the payment_asset for easier 1:1 conversions
- have a concrete example in `close_sales_at` function for calculations, refunds and contributor voucher redemption

## TODO
- verify the code and commits
- if agreement, merge to base PR branch
- next steps would be thinking about how we can manage the `SaleParticipants` or `Contributions` (rename the storage item)
  - we may need double storage map instead
  - we may need to limit no. of participants that can redeem vouchers per block

### DIFF
--- a/pallet/crowdsale/src/impls.rs
+++ b/pallet/crowdsale/src/impls.rs
@@ -14,16 +14,13 @@ use scale_info::prelude::format;
 
 impl<T: Config> Pallet<T> {
 	/// Creates a unique voucher asset for a sale. Returns the AssetId of the created asset
-	pub(crate) fn create_voucher_asset(
-		sale_id: SaleId,
-		decimals: u8,
-	) -> Result<AssetId, DispatchError> {
+	pub(crate) fn create_voucher_asset(sale_id: SaleId) -> Result<AssetId, DispatchError> {
 		let voucher_owner = T::PalletId::get().into_account_truncating();
 		let voucher_asset_id = T::MultiCurrency::create_with_metadata(
 			&voucher_owner,
 			format!("CrowdSale Voucher-{}", sale_id).as_bytes().to_vec(),
 			format!("CSV-{}", sale_id).as_bytes().to_vec(),
-			decimals,
+			VOUCHER_DECIMALS,
 			None,
 		)
 		.map_err(|_| Error::<T>::CreateAssetFailed)?;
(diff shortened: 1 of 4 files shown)


pallet/crowdsale/src/impls.rs
```diff
@@ -52,7 +55,8 @@ impl<T: Config> Pallet<T> {
 		};
 
 		// Return total vouchers converted to the correct decimals
-		account_contribution * 10u128.pow(VOUCHER_DECIMALS as u32) / voucher_price
+		todo!("Calculate voucher rewards based on the account contribution and voucher price.")
```
JasonTulp, line 58: Why this change? The calculations were working as expected?

pallet/crowdsale/src/impls.rs
```diff
@@ -14,13 +14,16 @@ use scale_info::prelude::format;
...
+		sale_id: SaleId,
+		decimals: u8,
+	) -> Result<AssetId, DispatchError> {
 		let voucher_owner = T::PalletId::get().into_account_truncating();
 		let voucher_asset_id = T::MultiCurrency::create_with_metadata(
 			&voucher_owner,
 			format!("CrowdSale Voucher-{}", sale_id).as_bytes().to_vec(),
 			format!("CSV-{}", sale_id).as_bytes().to_vec(),
-			VOUCHER_DECIMALS,
+			decimals,
```
JasonTulp, line 26: In the requirements it said that decimals for vouchers will always be 6DP (Hence why I moved it to a constant)

pallet/crowdsale/src/impls.rs
```diff
@@ -70,9 +74,58 @@ impl<T: Config> Pallet<T> {
...
+				let vouchers_quantity_redeemed = contribution / voucher_price; // 500_000_000 / 20_000_000 = 25
+				let voucher_decimals = T::MultiCurrency::decimals(&sale_info.payment_asset);
+				let voucher_amount = vouchers_quantity_redeemed
+					.saturating_mul(10u32.pow(voucher_decimals as u32).into());
+				T::MultiCurrency::mint_into(sale_info.voucher, &contributor, voucher_amount)
+					.map_err(|_| Error::<T>::AssetMintFailed)?;
+
+				// TODO: emit an event for each contributor redeeming their vouchers
+
+				// close the sale
```
JasonTulp, line 120: I don't think this should be in on_initialize.. We want to mark the sale as complete then use the offchain worker to calculate their voucher payout based on contribution

PR #794, opened 2024-02-26
Crowdsale voucher rewards calculations

JasonTulp:
Improves crowdsale voucher rewards calculations and eliminate the possibility of any rounding errors when performing calculations. 
Adds many unit tests to test different edge cases

### DIFF
--- a/pallet/assets-ext/src/lib.rs
+++ b/pallet/assets-ext/src/lib.rs
@@ -239,6 +239,9 @@ pub mod pallet {
 			let who = ensure_signed(origin)?;
 			// Decimals cannot be higher than 18 due to a restriction in the conversion function
 			// scale_wei_to_correct_decimals
+			// There is also a restriction with calculations in the crowdsale pallet where if
+			// an asset has higher than 29 decimal places, the calculations will return highly
+			// saturated results
 			ensure!(decimals <= MAX_DECIMALS, Error::<T>::DecimalsTooHigh);
 			// reserves some native currency from the user - as this should be a costly operation
 			let deposit = <AssetDeposit<T>>::get();
--- a/pallet/assets-ext/src/lib.rs
+++ b/pallet/assets-ext/src/lib.rs
@@ -226,6 +226,8 @@ pub mod pallet {
 		}
 
 		/// Creates a new asset with unique ID according to the network asset id scheme.
+		/// Decimals cannot be higher than 18 due to a restriction in the conversion function
+		/// scale_wei_to_correct_decimals
 		#[pallet::weight(< T as Config >::WeightInfo::create_asset())]
 		#[transactional]
 		pub fn create_asset(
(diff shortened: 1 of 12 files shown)


pallet/assets-ext/src/lib.rs
```diff
@@ -239,6 +239,9 @@ pub mod pallet {
 			let who = ensure_signed(origin)?;
 			// Decimals cannot be higher than 18 due to a restriction in the conversion function
 			// scale_wei_to_correct_decimals
+			// There is also a restriction with calculations in the crowdsale pallet where if
+			// an asset has higher than 29 decimal places, the calculations will return highly
+			// saturated results
 			ensure!(decimals <= MAX_DECIMALS, Error::<T>::DecimalsTooHigh);
```
zees-dev, line 245: we should just have this limitation when creating an asset in the first place.
I can imagine many pallets - even in the future where this restriction could apply; we should fix the source.
18 should probably be the max

pallet/assets-ext/src/lib.rs
```diff
@@ -239,6 +239,9 @@ pub mod pallet {
 			let who = ensure_signed(origin)?;
 			// Decimals cannot be higher than 18 due to a restriction in the conversion function
 			// scale_wei_to_correct_decimals
+			// There is also a restriction with calculations in the crowdsale pallet where if
+			// an asset has higher than 29 decimal places, the calculations will return highly
+			// saturated results
```
zees-dev, line 244: Move both these comments as 2 seperate bullet points to the doc comment above the function.
IMO this restriction is helpful to know about and should be displayed when generating Cargo docs.

pallet/crowdsale/src/impls.rs
```diff
@@ -44,26 +44,107 @@ impl<T: Config> Pallet<T> {
...
+		}
+
+		// Total contributions of all payments prior to this payment + the contributions
+		// from this account
+		// Converted to U256 to avoid overflow during calculations
+		let account_contribution = U256::from(account_contribution);
+		let total_paid_contributions = U256::from(total_paid_contributions);
+		let contribution_after: U256 =
+			account_contribution.saturating_add(total_paid_contributions);
+
```
zees-dev, line 126: rather than re-assigning the same vars to U256; maybe make all the function params be U256 - can simply `.into()` them when inputting in the function.

pallet/crowdsale/src/impls.rs
```diff
@@ -44,26 +44,107 @@ impl<T: Config> Pallet<T> {
...
+			// We are over committed. Calculate the voucher price based on the total
+			total_funds_raised.checked_div(voucher_max_supply).unwrap_or(0)
+		} else {
+			// We are under committed so we will pay out the soft cap
+			soft_cap_price
+		};
+		// Check prior to calculations to avoid division by zero
+		if voucher_price == 0 {
+			return 0
+		}
```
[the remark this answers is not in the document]
JasonTulp: Yep agreed, although we still need to check here to prevent dividing by zero

pallet/crowdsale/src/impls.rs
JasonTulp: Hmm I'm not too sure about this one, we only  need one part of the equation to have the accuracy of U256 (where we add 6 0s and multiply large numbers together) so there's no need to convert all the numbers to U256. Happy to discuss with you

zees-dev (approved):
LGTM 👍

PR #796, opened 2024-02-27
Crowdsale voucher payout + automatic payout

JasonTulp:
Adds manual voucher redeem + automatic voucher payout

TODOS:
- Fix build errors
- Tests tests tests
- Double check drain_prefix stuff. seems ok, but ok?
- Tests for the max_supply with the refunded vouchers removed. 
- Need more tests for calculate_voucher_rewards for when it is undercommitted. Using the refunded amount as the value within****

### DIFF
--- a/pallet/crowdsale/src/impls.rs
+++ b/pallet/crowdsale/src/impls.rs
@@ -139,7 +139,6 @@ impl<T: Config> Pallet<T> {
 			contribution_after.saturating_mul(U256::from(10_u128.pow(VOUCHER_DECIMALS as u32)));
 
 		// The total supply of vouchers after this payment is made
-		// Use checked div, if voucher_price is 0, return 0
 		let voucher_supply_after = contribution_after
 			.checked_div(U256::from(voucher_price))
 			.ok_or("Voucher price must be greater than 0")?;
@@ -152,10 +151,60 @@ impl<T: Config> Pallet<T> {
 			voucher_max_supply * 10_u128.pow(VOUCHER_DECIMALS as u32),
 		);
 
+		// Remove the refunded vouchers from the total supply, we only care about the vouchers
+		// that were redeemed
+		// TODO need tests for this
+		let refunded_vouchers = total_funds_raised.saturating_sub(crowd_sale_target);
+		let voucher_current_supply = voucher_current_supply.saturating_sub(refunded_vouchers);
+
 		// Return the number of vouchers to be paid out, which is the difference between
 		// the total supply after this payment and the total supply before this payment
 		return Ok(voucher_supply_after.saturating_sub(voucher_current_supply))
 	}
+
+	// // Mints vouchers into a users wallet and returns the amount minted
+	// pub fn mint_user_vouchers(
+	// 	who: &T::AccountId,
+	// 	sale_id: SaleId,
+	// 	sale_info: &mut SaleInformation<T::AccountId, T::BlockNumber>,
+	// 	contribution: Balance,
+	// 	voucher_max_supply: Balance,
+	// 	voucher_current_supply: Balance,
+	// 	total_paid_contributions: Balance,
+	// ) -> Result<Balance, DispatchError> {
+	// 	// calculate the claimable vouchers
+	// 	let claimable_vouchers = Self::calculate_voucher_rewards(
+	// 		sale_info.soft_cap_price,
+	// 		sale_info.funds_raised,
+	// 		contribution.into(),
+	// 		voucher_max_supply.into(),
+	// 		voucher_current_supply,
+	// 		total_paid_contributions.into(),
+	// 	)
+	// 	.map_err(|e| {
+	// 		log!(
+	// 			error,
+	// 			"⛔️ failed to calculate voucher rewards for user {:?} in sale {:?}: {:?}",
+	// 			who,
+	// 			id,
+	// 			e
+	// 		);
+	// 		Error::<T>::VoucherClaimFailed
+	// 	})?;
+	//
+	// 	// mint claimable vouchers to the user
+	// 	T::MultiCurrency::mint_into(sale_info.voucher, &who, claimable_vouchers)?;
+	//
+	// 	// Update total_contributions
+	// 	sale_info.status = SaleStatus::Enabled(total_paid_contributions.saturaticontribution);
+	//
+	// 	Self::deposit_event(Event::CrowdsaleVouchersClaimed {
+	// 		id: sale_id,
+	// 		who,
+	// 		amount: claimable_vouchers,
+	// 	});
+	// 	Ok(claimable_vouchers)
+	// }
 
 	/// Close all crowdsales that are scheduled to end this block
 	pub(crate) fn close_sales_at(now: T::BlockNumber) -> Result<u32, &'static str> {
@@ -187,7 +236,12 @@ impl<T: Config> Pallet<T> {
 				)?;
 
 				// TODO: use NFTExt to get the collection max issuance
-				let collection_max_issuance = 1000;
+
+				let collection_info =
+					T::NFTExt::get_collection_info(sale_info.reward_collection_id)?;
+				let Some(collection_max_issuance) = collection_info.max_issuance else {
+					return Err(Error::<T>::MaxIssuanceNotSet.into());
+				};
 				let crowd_sale_target = sale_info.soft_cap_price * collection_max_issuance;
 
 				let refunded_vouchers = sale_info.funds_raised.saturating_sub(crowd_sale_target);
@@ -199,13 +253,26 @@ impl<T: Config> Pallet<T> {
 					)?;
 				}
 
-				// close the sale
-				sale_info.status = SaleStatus::Closed(now);
-
+				// Throw event to mark end of crowdsale
 				Self::deposit_event(Event::CrowdsaleClosed {
 					id: sale_id,
 					info: sale_info.clone(),
 				});
+
+				// Try append to distributingSales, if this fails due to upper vec bounds
+				// Set status to DistributionFailed and log the error
+				DistributingSales::<T>::try_append(sale_id).map_err(|_| {
+					sale_info.status = SaleStatus::DistributionFailed(now);
+					log!(error, "⛔️ failed to mark sale {:?} for distribution", sale_id);
+					Error::<T>::DistributingSaleFailed
+				})?;
+
+				if sale_info.funds_raised.is_zero() {
+					sale_info.status = SaleStatus::Ended(now, Balance::default());
+				} else {
+					// Mark the sale for distribution
+					sale_info.status = SaleStatus::Distributing(now, Balance::default());
+				}
 
 				Ok(())
 			})?;
--- a/pallet/crowdsale/src/impls.rs
+++ b/pallet/crowdsale/src/impls.rs
@@ -15,7 +15,7 @@ use frame_support::sp_runtime::traits::{BlakeTwo256, Hash};
 use sp_core::U256;
 
 impl<T: Config> Pallet<T> {
-	/// Create a unique account (vault) to hold funds (payment_asset) for a crowdsale
+	/// Create a unique account (vault) to hold funds (payment_asset) for a crowdsale.
 	pub(crate) fn vault_account(nonce: SaleId) -> T::AccountId {
 		let seed: T::AccountId = T::PalletId::get().into_account_truncating();
 		let entropy = (seed, nonce).using_encoded(BlakeTwo256::hash);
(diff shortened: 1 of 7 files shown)


pallet/crowdsale/src/lib.rs
```diff
@@ -127,6 +136,11 @@ pub mod pallet {
 		OptionQuery,
 	>;
 
+	/// A list of all sales currently being distributed
+	#[pallet::storage]
+	pub type DistributingSales<T: Config> =
```
zees-dev, line 141: nit:
- `SaleDistribution`
- `SalesDistribution`
- `Distribution`
(top 2 pref just because all other storage mappings prefixed with `Sale`)

pallet/crowdsale/src/lib.rs
```diff
@@ -189,12 +195,14 @@ pub mod pallet {
...
 		/// The NFT collection must not contain any minted NFTs
 		CollectionIssuanceNotZero,
-		/// Cannot manually start the sale as an automatic start block is set
-		ManualStartDisabled,
 		/// There are too many sales queued for this block, try again on a different block
 		TooManySales,
 		/// Vouchers have already been claimed
 		VouchersAlreadyClaimed,
+		/// Automatic trigger of sales distribution has failed
+		DistributingSaleFailed,
```
zees-dev, line 203: nit:
```suggestion
		SaleDistributionFailed,
```

pallet/crowdsale/src/lib.rs
```diff
@@ -305,6 +310,9 @@ pub mod pallet {
 			// ensure soft_cap_price is not zero - prevent future div by zero
 			ensure!(!soft_cap_price.is_zero(), Error::<T>::InvalidAsset);
 
+			// TODO
```
zees-dev, line 313: todo is addressed no?

pallet/crowdsale/src/impls.rs
```diff
@@ -199,14 +253,27 @@ impl<T: Config> Pallet<T> {
 					)?;
 				}
 
-				// close the sale
-				sale_info.status = SaleStatus::Closed(now);
-
+				// Throw event to mark end of crowdsale
```
zees-dev, line 256: nit:
```suggestion
				// Emit event to mark end of crowdsale
```

pallet/crowdsale/src/impls.rs
```diff
@@ -187,7 +236,12 @@ impl<T: Config> Pallet<T> {
 				)?;
 
 				// TODO: use NFTExt to get the collection max issuance
```
zees-dev, line 238: nit: can remove, had been addressed

pallet/crowdsale/src/impls.rs
```diff
@@ -187,7 +236,12 @@ impl<T: Config> Pallet<T> {
...
-				let collection_max_issuance = 1000;
+
+				let collection_info =
+					T::NFTExt::get_collection_info(sale_info.reward_collection_id)?;
+				let Some(collection_max_issuance) = collection_info.max_issuance else {
+					return Err(Error::<T>::MaxIssuanceNotSet.into());
+				};
```
zees-dev, line 244: can use .ok_or

pallet/crowdsale/src/lib.rs
```diff
@@ -470,24 +478,84 @@ pub mod pallet {
...
+			// Get the first sale_id and process in FiFo order
+			let sale_id = sale_ids.first().ok_or(Error::<T>::CrowdsaleNotFound)?;
+			let mut sale_info = SaleInfo::<T>::get(sale_id).ok_or(Error::<T>::CrowdsaleNotFound)?;
+
+			// ensure the sale is in the distribution phase
+			let SaleStatus::Distributing(_, total_paid_contributions) = sale_info.status else {
+				return Err(Error::<T>::InvalidCrowdsaleStatus.into());
+			};
+
+			// get amount of claimable vouchers based on the user's contribution
```
zees-dev, line 495: comment does not reflect what's happening below

pallet/crowdsale/src/lib.rs
```diff
@@ -470,24 +478,84 @@ pub mod pallet {
...
+			let collection_info = T::NFTExt::get_collection_info(sale_info.reward_collection_id)?;
+			let voucher_max_supply =
+				collection_info.max_issuance.ok_or(Error::<T>::MaxIssuanceNotSet)?;
+
+			let mut voucher_current_supply = T::MultiCurrency::total_issuance(sale_info.voucher);
```
zees-dev, line 500: Just thinking, since admin get's minted voucher first, if they burn voucher - would this value get affected?

pallet/crowdsale/src/lib.rs
```diff
@@ -470,24 +478,84 @@ pub mod pallet {
...
+			// If the sale didn't reach the soft cap, then the admin will have some balance that
+			// is included in the currency::total_issuance
+			let crowd_sale_target = sale_info.soft_cap_price * voucher_max_supply.into();
+			let refunded_vouchers = sale_info.funds_raised.saturating_sub(crowd_sale_target);
+			voucher_current_supply = voucher_current_supply.saturating_sub(refunded_vouchers);
```
zees-dev, line 505: nit: Hmm maybe we should change this var name to `vouchers_distributed` (excluding admin payout)?

pallet/crowdsale/src/lib.rs
```diff
@@ -614,5 +705,46 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		/// In the very unlikely case that a sale was blocked from automatic distribution within
+		/// the on_initialise step. This function allows a manual trigger of distribution
+		/// callable by the admin of the sale
+		#[pallet::weight(0)]
```
zees-dev, line 712: why not [transactional]?

pallet/crowdsale/src/impls.rs
```diff
@@ -199,14 +253,27 @@ impl<T: Config> Pallet<T> {
...
 					info: sale_info.clone(),
 				});
 
+				// Try append to distributingSales, if this fails due to upper vec bounds
+				// Set status to DistributionFailed and log the error
+				DistributingSales::<T>::try_append(sale_id).map_err(|_| {
+					sale_info.status = SaleStatus::DistributionFailed(now);
+					log!(error, "⛔️ failed to mark sale {:?} for distribution", sale_id);
+					Error::<T>::DistributingSaleFailed
+				})?;
```
zees-dev, line 268: We can't propagate the error up (`?`) since the try_mutate will fail (and hence the sale will never close since end block will pass)

pallet/crowdsale/src/impls.rs
```diff
@@ -199,14 +253,27 @@ impl<T: Config> Pallet<T> {
...
 				}
 
-				// close the sale
-				sale_info.status = SaleStatus::Closed(now);
-
+				// Throw event to mark end of crowdsale
 				Self::deposit_event(Event::CrowdsaleClosed {
 					id: sale_id,
 					info: sale_info.clone(),
 				});
```
zees-dev, line 260: nit: move event to the bottom?

pallet/crowdsale/src/lib.rs
```diff
@@ -614,5 +705,46 @@ pub mod pallet {
...
+					sale_info.status = SaleStatus::Ended(block_number, Balance::default());
+				} else {
+					// Mark the sale for distribution
+					sale_info.status = SaleStatus::Distributing(block_number, Balance::default());
+				}
+
+				Self::deposit_event(Event::CrowdsaleClosed {
+					id: sale_id,
+					info: sale_info.clone(),
+				});
```
zees-dev, line 742: event is emitted already in `on_initialize` for this sale.
Maybe we need another event here?:
- `CrowdsaleManualDistribution`
- `CrowdsaleDistributionInitiated`
- `CrowdsaleForceDistribution`
etc.

pallet/crowdsale/src/lib.rs
```diff
@@ -574,17 +660,22 @@ pub mod pallet {
 		/// Emits `CrowdsaleNFTRedeemed` event when successful.
 		#[pallet::weight(0)]
 		#[transactional]
-		pub fn redeem(origin: OriginFor<T>, id: SaleId, quantity: TokenCount) -> DispatchResult {
+		pub fn redeem_voucher(
```
zees-dev, line 663: nice catch; confused with what we are redeeming haha

pallet/crowdsale/src/lib.rs
```diff
@@ -503,50 +571,52 @@ pub mod pallet {
...
 
 				// get amount of claimable vouchers based on the user's contribution
 				let collection_info =
 					T::NFTExt::get_collection_info(sale_info.reward_collection_id)?;
-				let voucher_total_supply =
+				let voucher_max_supply =
 					collection_info.max_issuance.ok_or(Error::<T>::MaxIssuanceNotSet)?;
 
+				let mut voucher_current_supply =
+					T::MultiCurrency::total_issuance(sale_info.voucher);
```
zees-dev, line 599: Im assuming if admin burns his vouchers for example then the math won't be affected?

pallet/crowdsale/src/lib.rs
```diff
@@ -557,6 +627,21 @@ pub mod pallet {
 					amount: claimable_vouchers,
 				});
 
+				let voucher_max_supply: Balance =
+					voucher_max_supply as u128 * 10u128.pow(VOUCHER_DECIMALS as u32);
+				let block_number = <frame_system::Pallet<T>>::block_number();
+				if voucher_current_supply + claimable_vouchers >= voucher_max_supply {
+					// Distribution complete
+					sale_info.status = SaleStatus::Ended(block_number, voucher_max_supply);
+					Self::deposit_event(Event::CrowdsaleDistributionComplete { id });
```
zees-dev, line 636: do you think it's worthwhile specifying here (as a param) how much was distributed?

pallet/crowdsale/src/lib.rs
```diff
@@ -614,5 +705,46 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		/// In the very unlikely case that a sale was blocked from automatic distribution within
+		/// the on_initialise step. This function allows a manual trigger of distribution
+		/// callable by the admin of the sale
+		#[pallet::weight(0)]
+		pub fn claim_blocked_sale(origin: OriginFor<T>, sale_id: SaleId) -> DispatchResult {
```
zees-dev, line 713: need to rename imo (naming is hard haha);
It's not a claim, moreso kickstarting a distribution.
Few things names I can think of:
- `try_distribution`
- `force_distribution`
- `initiate_distribution`
- `reset_distribution`
- `re_distribute`
etc.

pallet/crowdsale/src/lib.rs
JasonTulp: Good point, have addressed this by just adding the sum of total paid out vouchers to the enum :+1: Easier than trying to account for all edge cases of the admin trading their share

pallet/crowdsale/src/lib.rs
JasonTulp: Extrinsics are transactional by default, however I still use the tag on functions that can potentially error after changing state (To be overly cautious) 
This extrinsic doesn't do that though as the state is only changed after all potential failures

pallet/crowdsale/src/lib.rs
```diff
@@ -614,5 +705,46 @@ pub mod pallet {
...
+					return Err(Error::<T>::CrowdsaleNotFound.into());
+				};
+
+				// ensure the sale is in the correct state
+				ensure!(
+					matches!(sale_info.status, SaleStatus::DistributionFailed(_)),
+					Error::<T>::InvalidCrowdsaleStatus
+				);
+
+				ensure!(sale_info.admin == who, Error::<T>::AccessDenied);
```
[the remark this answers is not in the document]
JasonTulp: Yeah could do!

pallet/crowdsale/src/impls.rs
JasonTulp: Good catch! Fixed

pallet/crowdsale/src/impls.rs
JasonTulp: If we move the event to the bottom it won't be emitted  if we hit the failure case underneath

pallet/crowdsale/src/lib.rs
JasonTulp: Nice, that's a good idea. Will add the signer as a parameter so we can see who triggered it (After removing admin only permissions)

pallet/crowdsale/src/lib.rs
JasonTulp: Went with try_force_distribution

pallet/crowdsale/src/lib.rs
zees-dev: O nice, didn't know; good to know!

PR #798, opened 2024-02-28
cargo doc generation dispatch workflow

zees-dev:


JasonTulp (approved):
Damn nice, this will be useful :+1:

.github/workflows/generate-docs.yml
```diff
@@ -0,0 +1,46 @@
...
+        ref: ${{ github.event.inputs.branch }}
+        fetch-depth: 0
+
+    - uses: dtolnay/rust-toolchain@1.69
+
+    - name: Install toolchain
+      run: rustup show
+
+    - name: Generate Documentation
+      run: cargo doc -p ${{ github.event.inputs.pallet_name }}
```
surangap, line 37: might need to correct the input name. also might have to think about having multiple pallet names at once?

surangap (commented):
would be useful.

.github/workflows/generate-docs.yml
zees-dev: will iterate on it once it's merged; just wanted quick win for automating this

PR #801, opened 2024-03-01
Emit an event to know withdraw source for EBD

KarishmaBothara:
Bridging from TRN->ETH side, when it is a futureverse (proxyExtrinsic). EBD stores the sender (EOA) as the from address.
There is a requirement to use fpass address in these case. The PR will address this.

Also for XBD, we can track the event (xrplBridge.WithdrawRequestRequest)

Screenshot 2024-03-01 at 2 14 59 PM
which can give the sender(fpass address) associated with the extrinsic


JasonTulp (changes requested):
Why the need for a separate event? Can we not add a field to Erc20Withdraw event (emitted in `process_withdrawal`)

KarishmaBothara:
> Erc20Withdraw

Cause we have different function to process delayed withdrawal and normal withdrawal, which would mean we need to make update two events

```
Self::deposit_event(Event::<T>::Erc20Withdraw(
			asset_id,
			withdrawal_message.amount.saturated_into(),
			withdrawal_message.beneficiary,
		)); 
```

and 

```
if Self::process_withdrawal(withdrawal_message.clone(), asset_id).is_err() {
							Self::deposit_event(Event::<T>::DelayedErc20WithdrawalFailed(
								asset_id,
								withdrawal_message.beneficiary.into(),
							));
						}
```
and in process_delayed_payment, we don't know the source...

JasonTulp:
> > Erc20Withdraw
> 
> Cause we have different function to process delayed withdrawal and normal withdrawal, which would mean we need to make update two events
> 
> ```
> Self::deposit_event(Event::<T>::Erc20Withdraw(
> 			asset_id,
> 			withdrawal_message.amount.saturated_into(),
> 			withdrawal_message.beneficiary,
> 		)); 
> ```
> 
> and
> 
> ```
> if Self::process_withdrawal(withdrawal_message.clone(), asset_id).is_err() {
> 							Self::deposit_event(Event::<T>::DelayedErc20WithdrawalFailed(
> 								asset_id,
> 								withdrawal_message.beneficiary.into(),
> 							));
> 						}
> ```
> 
> and in process_delayed_payment, we don't know the source...

I see, that makes it tricky. We could potentially change the storage of a delayed withdrawal to include the sender address. Not entirely sure whether it's worth the hassle though. Easiest approach might be best (The separate events)

JasonTulp (commented):
Looks good Karishma, just need a storage migration for PendingPayments (As we have changed the enum variant)

PR #803, opened 2024-03-04
Pallet crowdsale benchmarks

zees-dev:
### Crowdsale pallet benchmarks

```sh
cargo test -p pallet-crowdsale --features="try-runtime,runtime-benchmarks" test_benchmarks
```


pallet/crowdsale/src/benchmarking.rs
```diff
@@ -0,0 +1,232 @@
...
+		let collection_id = T::NFTExt::do_create_collection(
+			acc.clone(),
+			BoundedVec::truncate_from("Hello".encode()),
+			0,
+			Some(1000),
+			None,
+			MetadataScheme::try_from(b"https://google.com/".as_slice()).unwrap(),
+			None,
+			OriginChain::Root,
+		).unwrap();
```
JasonTulp, line 113: These should be pulled out into a helper function to prevent duplicate code and simplify benchmarks

JasonTulp (approved):
lgtm! Nice work :)

PR #815, opened 2024-03-12
Crowdsale pallet fixes

zees-dev:
### Various fixes for crowdsale pallet

- Crowdsale collection can be public minted (if set) even though ownership has been transferred to vault.

### Updates
- Introduced `proxy_vault_call` extrinsic to crowdsale pallet
  - Allows sale admin to update crowdsale NFT collection metadata (name, base_uri)
- Updated `initialize` extrinsic in crowdsale pallet
  - Introduced additional optional params to set voucher asset name and symbol


pallet/crowdsale/src/lib.rs
```diff
@@ -348,6 +350,11 @@ pub mod pallet {
 			let max_issuance = max_issuance.ok_or(Error::<T>::MaxIssuanceNotSet)?;
 			ensure!(collection_issuance.is_zero(), Error::<T>::CollectionIssuanceNotZero);
 
+			// Verify collection public mint is disabled
+			if let Ok(mint_info) = T::NFTExt::get_public_mint_info(collection_id) {
```
JasonTulp, line 354: Maybe we should just use getCollectionInfo to optimise reads (Above we are reading collection_info for max_issuance and collection_issuance)
Another option is exposing a function such as "verify_crowdsale_eligibility" and perform these checks in NFT pallet

JasonTulp (commented):
Good catch!

pallet/crowdsale/src/lib.rs
zees-dev: Good point; need to think of an approach.
- In the first case, the `CollectionInfo` struct will need to be moved to the common pallet
- In the 2nd case we introduce tight coupling in NFT pallet

Another approach is exposing a validate function - i.e. `CollectionValidator` in the runtime (impls), and injecting this trait into crowdsale pallet config type (like `ExtrinsicChecker`).

zees-dev: According to the latest benchmark the weight is ~1-1.5% more for the `initialize` extrinsic - with the provided change/fix.
commit below: https://github.com/futureversecom/trn-seed/pull/815/commits/383748b87bee5a8ea3ef9f530438aacd3bc45d86

Given this; I don't think it's worthwhile re-structering the code/logic; would you agree?

JasonTulp: Actually this is my bad, it looks like get_public_mint_info doesn't read CollectionInfo so my comment is invalid, it reads a separate storage map

JasonTulp (approved):
lgtm!

surangap (approved):

PR #816, opened 2024-03-12
Add a toggle to enable / disable the delay mechanism for either deposit or withdraw + add assetId in events

KarishmaBothara:
- Add a toggle to enable / disable the delay mechanism for either deposit or withdraw
- Add assetId as the new argument when emitting delay event
- Add sender as the new argument when emitting delay event


JasonTulp (commented):
Looking good, however we need some tests against the 2 new extrinsics and the delay

pallet/erc20-peg/src/lib.rs
```diff
@@ -237,6 +241,20 @@ decl_module! {
 			WithdrawalsActive::put(activate);
 		}
 
+		/// Activate/deactivate delay deposits (root only)
+		#[weight = T::WeightInfo::activate_deposits()]
+		pub fn activate_deposits_delay(origin, activate: bool) {
+			ensure_root(origin)?;
+			DepositsDelayActive::put(activate);
```
JasonTulp, line 248: Should throw an event for these two extrinsics

pallet/erc20-peg/src/lib.rs
```diff
@@ -237,6 +241,20 @@ decl_module! {
...
+		}
+
+		/// Activate/deactivate withdrawals (root only)
+		#[weight = T::WeightInfo::activate_withdrawals()]
```
JasonTulp, line 252: Nit. Would be good to have benchmarks specific to these extrinsics. (Although they are the same weight, it can be useful if we ever change them. i.e. add events or storage reads/writes

pallet/erc20-peg/src/lib.rs
```diff
@@ -315,22 +333,29 @@ impl<T: Config> Module<T> {
...
-					},
-					WithdrawCallOrigin::Evm => {
-						// EVM payment delays are not supported
-						Err(Error::<T>::EvmWithdrawalFailed.into())
-					},
+		if Self::withdrawals_delay_active() {
+			if let Some((min_amount, delay)) = payment_delay {
+				if min_amount <= amount {
+					return match call_origin {
+						WithdrawCallOrigin::Runtime => {
```
JasonTulp, line 340: Is there a way to reduce some of this nesting?

pallet/erc20-peg/src/lib.rs
KarishmaBothara: makes sense

pallet/erc20-peg/src/lib.rs
KarishmaBothara: couldn't find a way

JasonTulp (approved):
LGTM!

KarishmaBothara:
> LGTM!

have made changes to set default value to true for DepositsDelayActive and WithdrawalsDelayActive to true as suggested by @ken-futureverse as this is the current behaviour on the chain

JasonTulp (approved):
lgtm :)

PR #824, opened 2024-03-25
Pallet XRPL ed25519 TX signature support

zees-dev:
## Background

XRPL [officially supports](https://xrpl.org/docs/concepts/accounts/cryptographic-keys#signing-algorithms) transactions signed using both `secp256k1` and `ed25519` key types (`ECDSA` and `EdDSA` signing algorithms respectively).
The XRPL pallet currently only supports messages signed via `secp256k1` keys (`ECDSA`).

This PR adds support for messages signed via `ed25519` keys (`EdDSA` algorithm).

## Changelog

- Added support for XRPL transactions (signed messages) which are signed using the `ed25519` signature scheme/algorithm.
- Support for Ethereum address generation using `32` byte `ed25519` public key
  - EOA address derivation: last 20 bytes of `keccak256(public_key)`
- `transact` extrinsic benchmark updated
- `XRPLExtrinsicExecuted` event updated
- Test updates (unit + e2e)


e2e/test/XRPL.test.ts
```diff
@@ -160,6 +162,97 @@ describe("XRPL pallet", () => {
...
 
+  it("can submit system remark extrinsic - using ed25519 signature", async () => {
+    const importedAccount = AccountLib.derive.familySeed("sEdS4rAgVysUtD5Zmm9F8i8uJBGik4K");
+    const signerInstance = AccountLib.derive.privatekey(importedAccount.keypair.privateKey!);
+    const publicKey = computePublicKey(`0x${signerInstance.keypair.publicKey!}`, true);
+    const eoa = Web3.utils.toChecksumAddress(
+      // remove "ED" prefix from public key to compute EOA
+      // keccak hash produces 32 bytes (64 chars) - take last 20 bytes (40 chars)
+      // remove "0x" prefix from keccak hash output (2 chars)
+      // get last 20 bytes of the keccak hash output (12 bytes - 24 chars)
```
JasonTulp, line 173: This is interesting, is there any further documentation that you could link to that goes over the reason for this in more detail?

pallet/xrpl/src/types.rs
```diff
@@ -84,18 +114,32 @@ pub struct XRPLTransaction {
...
+		let key_prefix = &self.signing_pub_key[0..2].to_uppercase();
+		match key_prefix.as_str() {
+			"ED" => {
+				// remove the first byte (prefix) from the public key
+				let pub_key: [u8; 32] =
+					pub_key[1..].try_into().map_err(|_| "Invalid length of decoded bytes")?;
+				let public = ed25519::Public::from_raw(pub_key);
+				Ok(XrplPublicKey::ED25519(public))
+			},
+			_ => {
```
JasonTulp, line 131: Could there be a case where an ECDSA public key is prefixed with "ED" and is wrongly assigned as an ED25519 public key?

e2e/test/XRPL.test.ts
zees-dev: nope; there is no standard for generating an EOA using edd2519 algorithm.
Every chain does it in it's own way - since it's unsupported.

For TRN, Im just using the simplest implementation at the node level (so that the logic is very similar to eoa generation via ecdsa).
The logic for this in TS is a bit weird since we ethers compute address doesn't give the correct matching EOA (that's generated on the node).

JasonTulp (commented):
Looks good, tests are thorough. Although I worry that an ECDSA public key could be prefixed with ED which would cause transactions to fail as it would attempt to decode as an ED25519 key

pallet/xrpl/src/types.rs
JasonTulp: Sweet as! Thanks for clarifying, other than that looks good. Nice work

JasonTulp (approved):
lgtm!

surangap (approved):
lgtm. Hope there isn't a requirement for the user to control funds at the derived TRN address directly.

zees-dev:
> Member

Good point; if ever this is the case - we'd be mapping and deriving eoa accounts from ed25519 public keys using the same approach.

ken-futureverse:
let's audit this and we merge.

PR #826, opened 2024-04-02
Fix token unlock bug for auctions

JasonTulp:
Currently when an auction is closed with no bids, the tokens are not unlocked and remain in a locked state. This PR addresses this


pallet/marketplace/src/impls.rs
```diff
@@ -446,6 +446,7 @@ impl<T: Config> Pallet<T> {
 				},
 				Listing::Auction(listing) => {
 					Self::remove_listing(listing_outer, listing_id);
+					let _ = listing.tokens.unlock_tokens(&listing.seller);
```
zees-dev, line 449: hmm `process_auction_closure` also does an unlock too.
i.e. `process_auction_closure` -> `process_payment_and_transfer` -> `tokens.unlock_and_transfer`

This implies that you may be doing redundant storage access (double removal).

A better solution maybe to only unlock the tokens in the `process_auction_closure` function within in this block:
```rust
let Some((winner, hammer_price)) = winning_bid else {
  // unlock here instead
  // ...
  return
}
```

zees-dev (commented):
Everything else LGTM - just a small query WRT storage accesses.
You will also need to re-benchmark for any updated extrinsics.

pallet/marketplace/src/impls.rs
leo-plugdefi: agree here, as unlock should be part of the process_auction_closure

JasonTulp: Good call, overlooked this. Addressed here: https://github.com/futureversecom/trn-seed/pull/826/commits/02a7b217a6f99b3f0735f50ad6491919cade69ff

zees-dev (approved):
LGTM 👍

leo-plugdefi (approved):

leo-plugdefi (commented):
Looks like there is a conflict on migration files, and it seems like more folks are running into this snag right now. Hopefully, we can push it to the mainnet soon.

PR #827, opened 2024-04-04
Futurepass migration extrinsic removal

zees-dev:
## Changelog

- Removed futurepass migration extrinsic (deprecated and unused)
  - This should reduce the runtime WASM size for futurepass pallet
- Added migration to purge on-chain storage item (migrator admin account)
- Updated mocks and tests accordingly
- Fixed bug with futurepass extrinsics not being benchmarkable (build failed with `runtime-benchmarks` flag)


pallet/futurepass/src/lib.rs
```diff
@@ -178,10 +157,6 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type DefaultProxy<T: Config> = StorageMap<_, Twox64Concat, T::AccountId, T::AccountId>;
 
-	/// Migration data for user (root) and collections they can migrate
-	#[pallet::storage]
-	pub type MigrationAdmin<T: Config> = StorageValue<_, T::AccountId, OptionQuery>;
```
surangap, line 183: nit - remove the existing value. kill the storage manually or write a small cleanup migration.

surangap (approved):
lgtm

pallet/futurepass/src/lib.rs
zees-dev: Ah yep good point; will look into simply destroying the storage value without a migration.
Might need sudo access for that.

JasonTulp (approved):
Lgtm!

zees-dev:
@JasonTulp, @surangap - updated PR to incl. migration of storage item removal

runtime/src/migrations/futurepass.rs
```diff
@@ -0,0 +1,151 @@
...
+		let pallet_prefix = Twox128::hash(b"Futurepass");
+		let storage_item_name = Twox128::hash(b"MigrationAdmin");
+		let storage_key = [pallet_prefix, storage_item_name].concat();
+
+		// Remove the storage item
+		storage::clear(&storage_key);
+
+		log::info!(target: "Migration", "Futurepass: successfully killed MigrationAdmin storage item");
+
+		Weight::from_ref_time(1u64)
```
JasonTulp, line 110: Does clearing a storage count as a write in terms of weight?

JasonTulp (approved):
lgtm! Nice work

surangap (approved):

runtime/src/migrations/futurepass.rs
zees-dev: Yep I believe so (since you'd be nullifying it).

PR #829, opened 2024-04-04
Ethy pallet - Frame V2 Upgrade

JasonTulp:
Migrates ethy pallet to Frame V2


pallet/ethy/src/lib.rs
```diff
@@ -91,131 +91,36 @@ pub(crate) const LOG_TARGET: &str = "ethy";
...
-		build(|config: &GenesisConfig<T>| {
-			for new_signer in config.xrp_door_signers.iter() {
+
+	#[pallet::genesis_build]
+	impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
+		fn build(&self) {
+			for new_signer in self.xrp_door_signers.iter() {
 				XrplDoorSigners::<T>::insert(new_signer, true);
 			}
 			// set the NotaryXrplKeys as well
```
surangap, line 126: might not be the exactly correct way to update the `NotaryXrplKeys`. `NotaryKeys` might not be updated at this point. minor - can address later since no major impact.

pallet/ethy/src/lib.rs
```diff
@@ -225,62 +129,323 @@ decl_storage! {
...
-		RelayerBondDeposit(AccountId, Balance),
+		RelayerBondDeposit { relayer: T::AccountId, bond: Balance },
 		/// An account has withdrawn a relayer bond
-		RelayerBondWithdraw(AccountId, Balance),
+		RelayerBondWithdraw { relayer: T::AccountId, bond: Balance },
 		/// A new relayer has been set
-		RelayerSet(Option<AccountId>),
+		RelayerSet { relayer: Option<T::AccountId> },
 		/// Xrpl Door signers are set
 		XrplDoorSignersSet,
```
surangap, line 434: maybe let's publish the state of `XrplDoorSigners` with the event.

pallet/ethy/src/lib.rs
```diff
@@ -225,62 +129,323 @@ decl_storage! {
...
-		/// The schedule to unpause the bridge has failed (scheduled_block)
-		FinaliseScheduleFail(BlockNumber),
+		/// The schedule to unpause the bridge has failed
+		FinaliseScheduleFail { scheduled_block: T::BlockNumber },
 		/// The bridge contract address has been set
-		SetContractAddress(EthAddress),
+		SetContractAddress { address: EthAddress },
 		/// Xrpl authority set change request failed
-		XrplAuthoritySetChangeRequestFailed
+		XrplAuthoritySetChangeRequestFailed,
```
surangap, line 440: failure reason if possible

pallet/ethy/src/lib.rs
```diff
@@ -312,117 +477,188 @@ decl_error! {
...
 				// Change authority keys, we are 5 minutes before the next epoch
 				log!(trace, "💎 Epoch ends in 5 minutes, changing authorities");
 				Self::handle_authorities_change();
 			}
 
 			// 2) Process validated messages
 			// Removed message_id from MessagesValidAt and processes
-			let mut processed_message_ids = Self::processed_message_ids();
+			let mut processed_message_ids = ProcessedMessageIds::<T>::get();
 			for message_id in MessagesValidAt::<T>::take(block_number) {
```
surangap, line 507: might need some careful planning here. we need to make sure `MessagesValidAt` max bound is within the allocated weight limit for `on_initialize`. The usual norm is to allocate 10% of block weight for `on_initialize`. means for all pallets that make use of `on_initialize`. need some discussion around how we want to divide that total so that we can derive max bounds such as this.
This is an existing problem of the old pallet. we should address it if possible to address within a reasonable time.
Then need a good benchmark for `on_initialize`. open for disussion.

pallet/ethy/src/lib.rs
```diff
@@ -312,117 +477,188 @@ decl_error! {
...
-				ProcessedMessageIds::put(processed_message_ids);
+				ProcessedMessageIds::<T>::put(processed_message_ids);
 			}
 
 			// 3) Try process delayed proofs
 			consumed_weight += DbWeight::get().reads(2u64);
-			if PendingEventProofs::iter().next().is_some() && !Self::bridge_paused() {
-				let max_delayed_events = Self::delayed_event_proofs_per_block();
+			if PendingEventProofs::<T>::iter().next().is_some() && !BridgePaused::<T>::get() {
+				let max_delayed_events = DelayedEventProofsPerBlock::<T>::get();
```
surangap, line 566: what do you think about moving `DelayedEventProofsPerBlock` to a type alias in the config?

pallet/ethy/src/lib.rs
```diff
@@ -312,117 +477,188 @@ decl_error! {
...
 				// mark as processed
 				if let Err(idx) = processed_message_ids.binary_search(&message_id) {
 					processed_message_ids.insert(idx, message_id);
 				}
 				// Tidy up status check
-				PendingClaimStatus::remove(message_id);
+				PendingClaimStatus::<T>::remove(message_id);
 			}
 			if !processed_message_ids.is_empty() {
 				impls::prune_claim_ids(&mut processed_message_ids);
```
surangap, line 559: What do you think moving both prunning and PendingEventProofs processing to `on_idle`? should remove some heat out of `on_initialize`

pallet/ethy/src/lib.rs
```diff
@@ -456,112 +695,177 @@ decl_module! {
...
+			{
 				let event_id: EventClaimId = (*event_id).saturated_into();
-				ensure!(!PendingEventClaims::contains_key(event_id), Error::<T>::EventReplayPending); // NOTE(surangap): prune PendingEventClaims also?
-				if !Self::processed_message_ids().is_empty() {
-					ensure!( event_id > Self::processed_message_ids()[0] &&
-						Self::processed_message_ids().binary_search(&event_id).is_err() , Error::<T>::EventReplayProcessed);
+				ensure!(
+					!PendingEventClaims::<T>::contains_key(event_id),
+					Error::<T>::EventReplayPending
+				); // NOTE(surangap): prune PendingEventClaims also?
```
surangap, line 807:
```suggestion
				);
```

pallet/ethy/src/lib.rs
```diff
@@ -456,112 +695,177 @@ decl_module! {
...
-				let process_at: T::BlockNumber = <frame_system::Pallet<T>>::block_number() + Self::challenge_period();
-				<MessagesValidAt<T>>::append(process_at, event_id);
-
-				Self::deposit_event(Event::<T>::EventSubmit(event_id, event_claim, process_at));
+				let process_at: T::BlockNumber =
+					<frame_system::Pallet<T>>::block_number() + ChallengePeriod::<T>::get();
+				MessagesValidAt::<T>::mutate(process_at.clone(), |v| {
+					let mut message_ids = v.clone().into_inner();
+					message_ids.push(event_id);
+					let message_ids_bounded = WeakBoundedVec::force_from(
```
surangap, line 833: should we stop adding more if the bound is reached here? keep adding means we are adding extra pressure(unbounded) to `on_initialize()`. maybe we can add it to the next available block number?

pallet/ethy/src/lib.rs
```diff
@@ -456,112 +695,177 @@ decl_module! {
...
+			Ok(())
 		}
 
-		#[weight = DbWeight::get().writes(1) + DbWeight::get().reads(2)]
 		/// Submit a challenge for an event
 		/// Challenged events won't be processed until verified by validators
 		/// An event can only be challenged once
-		pub fn submit_challenge(origin, event_claim_id: EventClaimId) -> DispatchResult {
+		#[pallet::weight(DbWeight::get().writes(1) + DbWeight::get().reads(2))]
+		pub fn submit_challenge(
```
surangap, line 856: should be transactional
```suggestion
                #[transactional]
		pub fn submit_challenge(
```

runtime/src/lib.rs
```diff
@@ -978,6 +978,11 @@ parameter_types! {
 	pub const MaxNewSigners: u8 = 20;
 	/// 75 blocks is 5 minutes before the end of the era
 	pub const AuthorityChangeDelay: BlockNumber = 75_u32;
+
+	pub const MaxEthData: u32 = 1024;
+	pub const MaxChallenges: u32 = 100;
+	pub const MaxMessagesPerBlock: u32 = 1000;
+	pub const MaxCallRequests: u32 = 1000;
```
surangap, line 985: let's have a discussion for proper values for these as well.

pallet/ethy/src/lib.rs
JasonTulp: This is the same as how it was before in add_extra_genesis

pallet/ethy/src/lib.rs
JasonTulp: I agree this is a good idea, but maybe cleaner to address in follow up PR? That way we can separate out the necessary changes for Frame V2 and further refactoring work

pallet/ethy/src/lib.rs
JasonTulp: I'm happy to do this, although we would lose the ability to change it at runtime. But I also don't see a massive need for this as it will most likely remain constant

pallet/ethy/src/lib.rs
JasonTulp: The problem with adding it to the next block number is that there's a chance the next block number is full, then you have to keep iterating through block numbers until you find one that has room. I think this should be fine for now, but I'm keen to follow up with an on_idle solution that will safely handle overflow and never exceed block weight

pallet/ethy/src/lib.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/829/commits/d750cd8d1d501297c5609e16b89490515539026d

pallet/ethy/src/lib.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/829/commits/d750cd8d1d501297c5609e16b89490515539026d

pallet/ethy/src/lib.rs
```diff
@@ -225,62 +129,323 @@ decl_storage! {
...
+		StorageMap<_, Twox64Concat, EthCallId, BTreeMap<CheckedEthCallResult, u32>, OptionQuery>;
+
+	/// EthCallOracle request info
+	#[pallet::storage]
+	pub type EthCallRequestInfo<T: Config> =
+		StorageMap<_, Twox64Concat, EthCallId, CheckedEthCallRequest<T::MaxEthData>, OptionQuery>;
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
```
zees-dev, line 395: The events are essentially being changed here; ensure downstream dependencies/services are aware

pallet/ethy/src/lib.rs
```diff
@@ -312,117 +477,188 @@ decl_error! {
...
 			}
 
 			consumed_weight
 		}
 
-		#[weight = DbWeight::get().writes(new_signers.len() as u64).saturating_add(DbWeight::get().reads_writes(4, 3))]
+		fn offchain_worker(block_number: T::BlockNumber) {
+			let active_notaries = NotaryKeys::<T>::get().into_inner();
+			log!(trace, "💎 entering off-chain worker: {:?}", block_number);
+			log!(trace, "💎 active notaries: {:?}", active_notaries);
```
zees-dev, line 584: lets upgrade these to `debug` log levels; in most cases i'd imagine we'd be explicitly debugging for offchain worker flows

pallet/ethy/src/lib.rs
```diff
@@ -456,112 +695,178 @@ decl_module! {
...
 			Ok(())
 		}
 
-		#[weight = DbWeight::get().writes(1)]
-		/// Set event confirmations (blocks). Required block confirmations for an Ethereum event to be notarized by Seed
-		pub fn set_event_block_confirmations(origin, confirmations: u64) {
+		/// Set event confirmations (blocks). Required block confirmations for an Ethereum event to
+		/// be notarized by Seed
+		#[pallet::weight(DbWeight::get().writes(1))]
+		pub fn set_event_block_confirmations(
```
zees-dev, line 710: nit: emit event here

pallet/ethy/src/lib.rs
```diff
@@ -456,112 +695,178 @@ decl_module! {
...
-			EventBlockConfirmations::put(confirmations)
+			EventBlockConfirmations::<T>::put(confirmations);
+			Ok(())
 		}
 
-		#[weight = DbWeight::get().writes(1)]
 		/// Set max number of delayed events that can be processed per block
-		pub fn set_delayed_event_proofs_per_block(origin, count: u8) {
+		#[pallet::weight(DbWeight::get().writes(1))]
+		pub fn set_delayed_event_proofs_per_block(
```
zees-dev, line 721: nit: emit event here

pallet/ethy/src/lib.rs
```diff
@@ -456,112 +695,178 @@ decl_module! {
...
+			Ok(())
 		}
 
-		#[weight = DbWeight::get().writes(1)]
-		/// Set challenge period, this is the window in which an event can be challenged before processing
-		pub fn set_challenge_period(origin, blocks: T::BlockNumber) {
+		/// Set challenge period, this is the window in which an event can be challenged before
+		/// processing
+		#[pallet::weight(DbWeight::get().writes(1))]
+		pub fn set_challenge_period(
```
zees-dev, line 733: nit: emit event here

pallet/ethy/src/lib.rs
```diff
@@ -456,112 +695,178 @@ decl_module! {
...
+			ContractAddress::<T>::put(contract_address);
+			Self::deposit_event(Event::<T>::SetContractAddress { address: contract_address });
+			Ok(())
 		}
 
-		#[weight = DbWeight::get().writes(1)]
 		/// Pause or unpause the bridge (requires governance)
-		pub fn set_bridge_paused(origin, paused: bool) {
+		#[pallet::weight(DbWeight::get().writes(1))]
+		pub fn set_bridge_paused(origin: OriginFor<T>, paused: bool) -> DispatchResult {
```
zees-dev, line 756: nit: emit event here

pallet/ethy/src/lib.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/829/commits/5f6d834c299c1adc79d5ad9268f244ccfe2a1d54

pallet/ethy/src/lib.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/829/commits/5f6d834c299c1adc79d5ad9268f244ccfe2a1d54

pallet/ethy/src/lib.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/829/commits/5f6d834c299c1adc79d5ad9268f244ccfe2a1d54

pallet/ethy/src/lib.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/829/commits/5f6d834c299c1adc79d5ad9268f244ccfe2a1d54

runtime/src/lib.rs
```diff
@@ -978,6 +978,11 @@ parameter_types! {
 	pub const MaxNewSigners: u8 = 20;
 	/// 75 blocks is 5 minutes before the end of the era
 	pub const AuthorityChangeDelay: BlockNumber = 75_u32;
+
+	pub const MaxEthData: u32 = 3200;
+	pub const MaxChallenges: u32 = 100;
+	pub const MaxMessagesPerBlock: u32 = 1000;
```
surangap, line 984: `MaxMessagesPerBlock` and `MaxEthData` might need further optimizations. will be considered in a later PR

surangap (approved):
further improvement can be done in a later PR.
lgtm, good work 👏

PR #831, opened 2024-04-14
Ethy on_initialise optimisations

JasonTulp:
This PR addresses some optimisation improvements for Ethy pallet and reduces some of the load in on_initialize. 

- Move delayed event processing to on_idle to remove some strain in on_initialize
- Update the ProcessedMessageId pruning algorithm to add an upper limit (1000) to the amount of Ids that are stored within this list. Previously the list is ever growing and iterated through every single block which is bad practice and could become a problem. 
- Prevent pruning of ProcessedMessageIds unless required (Instead of every block) 
- Adds submit_missing_event extrinsic which allows for submitting events that have been pruned by the more aggressive pruning algorithm
- Adds admin extrinsic remove_missing_event_id which allows for sudo to remove a range of event_ids from the MissedMessageIds storage to clear space


zees-dev:
needs to be rebased (conflicts)

pallet/ethy/src/impls.rs
```diff
@@ -94,6 +94,111 @@ impl<T: Config> XrplBridgeToEthyAdapter<T::EthyId> for Pallet<T> {
...
 
 impl<T: Config> Pallet<T> {
+	/// Prunes claim ids that are less than the max contiguous claim id.
+	pub(crate) fn prune_claim_ids(claim_ids: &mut Vec<EventClaimId>) {
+		// if < 1 element, nothing to do
+		if let 0..=1 = claim_ids.len() {
+			return
+		}
+
+		// Take the last ProcessedMessageIdBuffer elements in the list
```
surangap, line 104:
```suggestion
		// Keep the last ProcessedMessageIdBuffer elements in the list
```

pallet/ethy/src/lib.rs
```diff
@@ -334,6 +348,11 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type ProcessedMessageIds<T> = StorageValue<_, Vec<EventClaimId>, ValueQuery>;
 
+	/// Tracks message Ids that are outside of the MessageId buffer and were not processed
+	/// These message Ids can be either processed or cleared by the relayer
+	#[pallet::storage]
+	pub type MissedMessageIds<T> = StorageValue<_, Vec<EventClaimId>, ValueQuery>;
```
surangap, line 354: might be able to put a bound now since `ProcessedMessageIdBuffer` and get rid of `without_storage_info` for good?

pallet/ethy/src/lib.rs
```diff
@@ -500,18 +519,21 @@ pub mod pallet {
...
-			let mut consumed_weight = Weight::zero();
+			// Reads: NextAuthorityChange, MessagesValidAt, ProcessedMessageIds
+			let mut consumed_weight = DbWeight::get().reads(3u64);
 
 			// 1) Handle authority change
 			if Some(block_number) == NextAuthorityChange::<T>::get() {
 				// Change authority keys, we are 5 minutes before the next epoch
 				log!(trace, "💎 Epoch ends in 5 minutes, changing authorities");
+				// TODO Benchmark this function
 				Self::handle_authorities_change();
```
surangap, line 530: weight for this function. TODO above

pallet/ethy/src/lib.rs
```diff
@@ -556,31 +578,73 @@ pub mod pallet {
...
+
 				// Tidy up status check
 				PendingClaimStatus::<T>::remove(message_id);
+				message_processed = true;
 			}
-			if !processed_message_ids.is_empty() {
-				impls::prune_claim_ids(&mut processed_message_ids);
+
+			if message_processed && !processed_message_ids.is_empty() {
+				Self::prune_claim_ids(&mut processed_message_ids);
```
surangap, line 604: weight for this function

pallet/ethy/src/lib.rs
```diff
@@ -556,31 +578,73 @@ pub mod pallet {
...
+				return Weight::zero()
+			}
+
+			// Don't do anything if the bridge is paused
+			if BridgePaused::<T>::get() {
+				return DbWeight::get().reads(1u64)
+			}
+
+			let mut consumed_weight = base_weight;
+			let mut pending_event_proofs = PendingEventProofs::<T>::drain();
```
surangap, line 634: seems we drain the whole thing here. need to make sure unprocessed items are not being removed at the exit.

pallet/ethy/src/lib.rs
```diff
@@ -500,18 +519,21 @@ pub mod pallet {
...
+				// TODO Benchmark this function
 				Self::handle_authorities_change();
 			}
 
 			// 2) Process validated messages
 			// Removed message_id from MessagesValidAt and processes
 			let mut processed_message_ids = ProcessedMessageIds::<T>::get();
+			let mut message_processed: bool = false;
 			for message_id in MessagesValidAt::<T>::take(block_number) {
 				if PendingClaimStatus::<T>::get(message_id) == Some(EventClaimStatus::Challenged) {
```
zees-dev, line 538: another db eead for this?

pallet/ethy/src/lib.rs
```diff
@@ -786,85 +850,81 @@ pub mod pallet {
 			Ok(())
 		}
 
+		/// Admin function to manually remove an event_id from MissedMessageIds
+		/// This should only be used if the event_id is confirmed to be invalid
+		/// event_id_range is the lower and upper event_ids to clear (Both Inclusive)
+		#[pallet::weight(T::WeightInfo::remove_missing_event_id())]
+		pub fn remove_missing_event_id(
```
zees-dev, line 857: lets emit an event here

pallet/ethy/src/lib.rs
```diff
@@ -786,85 +850,81 @@ pub mod pallet {
...
 		/// Submit ABI encoded event data from the Ethereum bridge contract
+		/// Used to recover events that were pruned but not handled by the pruning algorithn,
+		/// Only events contained within MissedMessageIds can be processed here
 		/// - tx_hash The Ethereum transaction hash which triggered the event
 		/// - event ABI encoded bridge event
-		#[pallet::weight(DbWeight::get().writes(1))]
-		pub fn submit_event(origin: OriginFor<T>, tx_hash: H256, event: Vec<u8>) -> DispatchResult {
+		#[pallet::weight(T::WeightInfo::submit_missing_event())]
+		#[transactional]
+		pub fn submit_missing_event(
```
zees-dev, line 878: and here; in fact, since changes are being made in this pallet, lets ensure any all state changing extrinsics emit events?

zees-dev: this function does seem to be emitting event - in impl however.
nit: would prefer if event is emitted in the top-level extrinsic code since its easier for readability. Not a major issue though.

pallet/ethy/src/lib.rs
JasonTulp: There is a test for this, drain will return an iterator and items are only removed when calling .next()
If not all items are processed, they will remain in the storagemap

pallet/ethy/src/lib.rs
JasonTulp: I left the event within the impl function to reduce nested code, we have used this pattern in many places within the codebase

pallet/ethy/src/lib.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/831/commits/7db2366d266f626b13146adb43c188c420d4d16d

pallet/ethy/src/lib.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/831/commits/7db2366d266f626b13146adb43c188c420d4d16d

pallet/ethy/src/lib.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/831/commits/7db2366d266f626b13146adb43c188c420d4d16d

pallet/ethy/src/lib.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/831/commits/7db2366d266f626b13146adb43c188c420d4d16d

pallet/ethy/src/impls.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/831/commits/7db2366d266f626b13146adb43c188c420d4d16d

pallet/ethy/src/lib.rs
JasonTulp: I've added a bound to ProcessedMessageIds, however MissedMessageIds is still unbounded as it may grow indefinitely. Happy to discuss bounding this and either truncating or using WeakBoundedVec. What do you think?

pallet/ethy/src/lib.rs
zees-dev: personally prefer events to be at the extrinsic/tx/top-level as it's made very clear - where impl would just have the business logic.
But yes that would require a change in the impl.
Nevertheless, not a biggie

zees-dev (approved):
LGTM 👍

pallet/ethy/src/lib.rs
surangap: I do not expect `MissedMessageIds` to grow into a larger list in the real world scenario. Maybe keep the same bound as `ProcessedMessageIdBuffer` and, you can use either of the methods above. can truncate from the oldest missing Id also.

surangap (approved):
lgtm👍
https://github.com/futureversecom/trn-seed/pull/831#discussion_r1599285129

pallet/ethy/src/lib.rs
```diff
@@ -332,7 +346,13 @@ pub mod pallet {
 	/// Tracks processed message Ids (prevent replay)
 	/// Must remain unbounded as this list will grow indefinitely
 	#[pallet::storage]
-	pub type ProcessedMessageIds<T> = StorageValue<_, Vec<EventClaimId>, ValueQuery>;
+	pub type ProcessedMessageIds<T: Config> =
```
surangap, line 349: might need a migration for existing data > 1000 ?

JasonTulp: Good catch https://github.com/futureversecom/trn-seed/pull/831/commits/280ec0f5b6e408f9969876c38d77939f89757941

runtime/src/migrations/ethy.rs
```diff
@@ -0,0 +1,176 @@
...
+	where
+		AccountId: From<H160>,
+	{
+		log::info!(target: "Migration", "Ethy: migrating ProcessedMessageIds");
+		let mut weight = Weight::zero();
+
+		let mut processed_message_ids =
+			Value::unsafe_storage_get::<Vec<EventClaimId>>(b"EthBridge", b"ProcessedMessageIds")
+				.unwrap_or_default();
+		let prune_weight = pallet_ethy::Pallet::<T>::prune_claim_ids(&mut processed_message_ids);
```
surangap, line 95: `MissedMessageIds` might shoot up with old missing ids. might be better to get an idea about the number here wrt mainnet data. if that's > bound(MissedMessageIds) or too large, maybe we can truncate starting from the old.

JasonTulp: MissedMessageIds will only contain ~120 values on mainnet and less on Porcini

PR #833, opened 2024-04-17
Fix to dry out ethy old messages and improvements

surangap:
### Changelog
- We add `EVENT_INVALID_BELOW` to discard old messages lesser than that.
- remove duplicate gossips of witneses by validators
- fixes in gossip validator
- passive peers gossip only if they have received finality notification and hence metadata.


surangap:
Released with https://github.com/futureversecom/trn-seed/pull/834

PR #837, opened 2024-04-23
Ethy gadget finalized block number based Witness pruning.

surangap:
### Changelog
 - Add `block_number` to `Witness`
 - Active validators now include finalized block number(that the proof request was made) in the Witness message
 - Define `WINDOW_SIZE = 90` or the live window, proof needs to be generated within this window. approximately ~6mins.
 - Update `GossipValidator` to include a reference to `Backend`
 - Any incoming Witness gossip messages that are outside the live window would be discarded.
 - change `REBROADCAST_AFTER` to 3 mins, allowing a second rebroadcast within the live window
 - Increase `MAX_COMPLETE_EVENT_CACHE` to 500
 
What we can expect with this change
- All the existing gossip Witness messages would be discarded due to the new format
- There is a chance of discarding witnesses before they can form the proof if goes out of the live window. we need a separate mechanism to handle these scenarios. will be introduced later.

Note - at post deployment, we need to make sure  at least 66% of the active validators are upgraded to the new image. otherwise could halt the proof formation due to not being able to reach consensus.

### DIFF
--- a/ethy-gadget/src/witness_record.rs
+++ b/ethy-gadget/src/witness_record.rs
@@ -129,7 +129,7 @@ impl WitnessRecord {
 		}
 		.unwrap_or(0_usize);
 
-		trace!(target: "ethy", "💎 event {:?}, has # support: {:?}", event_id, witness_count);
+		trace!(target: "ethy", "💎 event {:?}, has # support: {:?}, proof threshold: {:?}", event_id, witness_count, proof_threshold);
 		witness_count >= proof_threshold
 	}
 
@@ -291,7 +291,7 @@ fn compact_sequence(completed_events: &mut [EventProofId]) -> &[EventProofId] {
 			watermark_idx = i + 1;
 			continue
 		} else {
-			break
+			break // Note - fix the algo
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


ethy-gadget/src/witness_record.rs
```diff
@@ -291,7 +291,7 @@ fn compact_sequence(completed_events: &mut [EventProofId]) -> &[EventProofId] {
 			watermark_idx = i + 1;
 			continue
 		} else {
-			break
+			break // Note - fix the algo
```
JasonTulp, line 294: What do you mean by this comment?

primitives/src/ethy.rs
```diff
@@ -159,6 +159,8 @@ pub struct Witness {
 	pub authority_id: AuthorityId,
 	/// ECDSA signature over `digest`
 	pub signature: AuthoritySignature,
+	/// proof requested block number
+	pub block_number: u64,
```
JasonTulp, line 163: Will this change mean we'd need to migrate the client-side storage for Witnesses?

surangap: No, we don't store the Witnesses into the DB.

ethy-gadget/src/witness_record.rs
surangap: pruning algorithm at this level is not perfect. can lead to a larger list if due to an incomplete event id. same as ethy prunning problem. We need to change the algo. no big harm for now.

leo-plugdefi (approved):

JasonTulp (approved):
lgtm, nice work

surangap:
release branch -> https://github.com/futureversecom/trn-seed/pull/838

PR #847, opened 2024-05-30
pallet-assets-ext-benchmarks

zees-dev:
**PR checklist**  

- [ ] Describe the added/changed/fixed/removed behaviour
- [ ] Tag related issue(s)
- [ ] Add appropriate unit tests and/or integration tests to the [integration test repository](https://github.com/futureversecom/seed-integration-tests)


JasonTulp:
I think adding custom benchmarks for AssetsExt is a good idea, but I personally don't think it will solve the discrepency issue with the gas costs. It looks like the costs have barely changed and there are some included cost updates to areas of the code that aren't affected by the AssetsExt benchmark change? (i.e. ERC721, ERC1155, Futurepass etc) Unless I am missing something, I'm not sure why those changes were included

Also if we were to add benchmarks to AssetsExt, we should really create separate benchmarks for an asset or a native ROOT token, determined within the weight macro to provide the correct cost for both. 
    - i.e. transfer_balances, transfer_assets benchmark etc
Otherwise there is essentially no advantage to just using the weight for the inner asset pallet benchmark as it just proxies the extrinsic directly with no added logic (Bar some marginal extra time)

zees-dev:
> I think adding custom benchmarks for AssetsExt is a good idea, but I personally don't think it will solve the discrepency issue with the gas costs. It looks like the costs have barely changed and there are some included cost updates to areas of the code that aren't affected by the AssetsExt benchmark change? (i.e. ERC721, ERC1155, Futurepass etc) Unless I am missing something, I'm not sure why those changes were included
> 
> Also if we were to add benchmarks to AssetsExt, we should really create separate benchmarks for an asset or a native ROOT token, determined within the weight macro to provide the correct cost for both. - i.e. transfer_balances, transfer_assets benchmark etc Otherwise there is essentially no advantage to just using the weight for the inner asset pallet benchmark as it just proxies the extrinsic directly with no added logic (Bar some marginal extra time)

I think the benchmarks are also based on which machine they are run on; hence they've been normalized on my machine.
I do agree with your point WRT marginal differences in the end result here.
It must be noted that the currently referenced substrate weights are external. So with additional substrate upgrades, we would be pointing to external weight values, which are not benchmarked against reference machine.

surangap:
> I think adding custom benchmarks for AssetsExt is a good idea, but I personally don't think it will solve the discrepency issue with the gas costs. It looks like the costs have barely changed and there are some included cost updates to areas of the code that aren't affected by the AssetsExt benchmark change? (i.e. ERC721, ERC1155, Futurepass etc) Unless I am missing something, I'm not sure why those changes were included
> 
> Also if we were to add benchmarks to AssetsExt, we should really create separate benchmarks for an asset or a native ROOT token, determined within the weight macro to provide the correct cost for both. - i.e. transfer_balances, transfer_assets benchmark etc Otherwise there is essentially no advantage to just using the weight for the inner asset pallet benchmark as it just proxies the extrinsic directly with no added logic (Bar some marginal extra time)

agree here. I think if we are to benchmark the `AssetsExt`, we should have separate benchmarks for `Assets` and `Balances` paths. and then use the correct weight function based on the asset id passed with.
Also we use weight files generated on our reference machine for both `Assets` and `Balances` now.

PR #851, opened 2024-06-14
Cargo workspace dependency migration

zees-dev:
# Context

Migrating all common crate deps to using cargo workspaces.

- In this PR, all the common crates were extracted (using `cargo tree` + custom bash script)
- All deps are imported with `default-features = false`, specific features must explicitly be enabled; this minimizes the output binary (and potentially wasm) file size
  - This implies all existing dep imports have been defaulted to `default-features = false` where possible
- For crates/pallets that used different versions of the same dependency/lib, the highest dep version is used
- Dependencies which are crate specific (only appear in single crate), are isolated to the single crate
- `pallet-vortex` pallet and crate name updated to `pallet-vortex-distribution` to align with other pallets
  - also addresses the edge case in benchmarking script for this pallet
- updated evm precompile file directory `futurepass_registrar` -> `futurepass-registrar` (use kebab-case for dir/file names)

## Misc. changes

- updated CI to `check` and `test` all crates and features
  - this is required since the default bin is set in root `Cargo.toml` (cargo test by default would only test the `client` crate)


ethy-gadget/Cargo.toml
```diff
@@ -1,52 +1,53 @@
 [package]
 name = "ethy-gadget"
 version = "0.1.2"
-authors = ["Parity Technologies <admin@parity.io>", "The Root Network Team"]
```
JasonTulp, line 4: Nit. Maybe for files where the author is different than the workspace (i.e. not just TRN team) we should keep as is?

evm-precompiles/utils/macro/tests/tests.rs
```diff
@@ -21,14 +22,17 @@ pub enum Action {
 
 #[test]
 fn test_keccak256() {
-	assert_eq!(&precompile_utils_macro::keccak256!(""), Keccak256::digest(b"").as_ref(),);
+	assert_eq!(
+		&precompile_utils_macro::keccak256!(""),
```
JasonTulp, line 26: Why did these tests need changing?

evm-precompiles/utils/macro/src/lib.rs
```diff
@@ -89,7 +89,7 @@ pub fn generate_function_selector(_: TokenStream, input: TokenStream) -> TokenSt
 		match variant.discriminant {
 			Some((_, Expr::Lit(ExprLit { lit, .. }))) =>
 				if let Lit::Str(lit_str) = lit {
-					let digest = Keccak256::digest(lit_str.value().as_ref());
+					let digest = Keccak256::digest(lit_str.value().as_bytes());
```
JasonTulp, line 92: Why was this change required?

pallet/dex/rpc/runtime-api/Cargo.toml
```diff
@@ -1,18 +1,19 @@
 [package]
 name = "pallet-dex-rpc-runtime-api"
 version = "0.0.1"
-authors = ["The Root Network Team"]
-edition = "2021"
-license = "Apache-2.0"
+authors.workspace = true
```
JasonTulp, line 4: Nit. While you're here it could be worth adding a description to the cargo.toml files that are missing it

pallet/vortex-distribution/Cargo.toml
```diff
@@ -0,0 +1,54 @@
+[package]
+name = "pallet-vortex-distribution"
```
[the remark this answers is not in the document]
JasonTulp: Although this may cause breaking changes to the service that distributes vortex tokens. Need to communicate but should be minor

JasonTulp (approved):
lgtm!

evm-precompiles/utils/macro/src/lib.rs
zees-dev: This crate referenced the `sha3` dep - which was version `0.8`.
The latest version of `sha3` used in this repo is version `0.10` (in root `Cargo.toml` now).
This dep update was a breaking change - hence needed to update the code and test accordingly to align the dep so the single highest version of `sha3` `0.10` could be used.

pallet/vortex-distribution/Cargo.toml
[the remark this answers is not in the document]
zees-dev: I think it's a good time to address this now - rather than later since there are other breaking changes.
However, if there is disagreement; I can drop the commit(s) which introduced this change.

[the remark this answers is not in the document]
JasonTulp: No I agree, let's get it more consistent with the others

ethy-gadget/Cargo.toml
zees-dev: https://github.com/futureversecom/trn-seed/pull/851/commits/80de5a2a5bdfb16492f1340fbcec2b823cbc9e9c

pallet/dex/rpc/runtime-api/Cargo.toml
zees-dev: I think for the main pallets its worthwhile; maybe not for these rpc and runtime api ones.
I cant envision these being used outside the context of TRN - as they first require a dependency on the core pallet itself.

pallet/vortex-distribution/Cargo.toml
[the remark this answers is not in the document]
zees-dev: not a breaking change since in the runtime the pallet is still called `VortexDistribution`

runtime/Cargo.toml
```diff
@@ -5,136 +5,136 @@ authors = ["The Root Network Team"]
...
 
 [dev-dependencies]
-ethereum = { version = "0.14.0", features = ["with-codec"]  }
-seed-client = { path = "../client" }
-ethabi = { version = "18.0.0" }
-frame-remote-externalities = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v1.0.0" }
-tokio = { version = "1.19.2", features = ["macros"] }
-substrate-test-runtime-client = { version = "2.0.0", git = "https://github.com/paritytech/substrate", branch = "polkadot-v1.0.0" }
+ethereum = { version = "0.14.0", default-features = true }
+seed-client = { workspace = true, default-features = true }
```
surangap, line 133: any reason why `default-features = true` is explicitly mentioned? it would be enabled by default since the packages already define a default feature set?

runtime/Cargo.toml
```diff
@@ -5,136 +5,136 @@ authors = ["The Root Network Team"]
 edition = "2021"
 
 [dependencies]
-codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive"] }
-log = { version = "0.4.17", default-features = false }
-scale-info = { version = "2.3.0", default-features = false, features = ["derive"] }
-serde = { version = "1.0.145", default-features = false, optional = true, features = ["derive"] }
+codec = { workspace = true }
+log = { workspace = true }
+scale-info = { workspace = true }
```
surangap, line 10: `features = ["derive"]` is not required anymore?

runtime/Cargo.toml
```diff
@@ -5,136 +5,136 @@ authors = ["The Root Network Team"]
...
 
 # EVM support
-fp-rpc = { default-features = false, git = "https://github.com/futureversecom/frontier", branch = "polkadot-v1.0.0-TRN" }
-fp-self-contained = { default-features = false, git = "https://github.com/futureversecom/frontier", branch = "polkadot-v1.0.0-TRN", features = ["serde"] }
-fp-evm = { default-features = false, git = "https://github.com/futureversecom/frontier", branch = "polkadot-v1.0.0-TRN" }
-evm = { version = "0.36.0", default-features = false, features = ["with-codec"] }
+fp-rpc = { workspace = true }
+fp-self-contained = { workspace = true, features = ["serde"] }
+fp-evm = { workspace = true }
+evm = { workspace = true }
```
surangap, line 59: `features = ["with-codec"]` is not required?

surangap (approved):
lgtm. thanks for working on this.
Added a few comments for clarifications.

runtime/Cargo.toml
zees-dev: Yeap; runtime seems to compile without needing this.

runtime/Cargo.toml
zees-dev: yeap, runtime seems to compile without needing this.

PR #857, opened 2024-07-02
Update Substrate deps for native support

JasonTulp:
Updates our Substrate dependencies to a custom branch so we can include upstream changes for bringing back native support in our Executor


rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly-2023-05-22"
+channel = "nightly"
```
zees-dev, line 2: we need to be explicit with the exact rust version being used

JasonTulp: Yeah agreed, I'm in the process of trying to find a proper version for this, we have some upstream deps that rely on a newer version than above, however Substrate has moved to just "nightly" as shown, so I need to backtrace and find a version that suits

zees-dev:
Since the deps are moved to root Cargo.toml, and we are changing the core substrate dep; can we try patching the deps instead?
I think this should be possible and also will be transitive for frontier too (won't require changes in frontier).

surangap (approved):
would be very useful👏

zees-dev (approved):
Great work, love that we can use latest nightly! 🚀

PR #858, opened 2024-07-03
Cost figures generated on standard spec

surangap:
### Changelog
- Cost figures generated on standard spec.


zees-dev (approved):
LGTM 👍 

Note: We should probably run cost validation tests on every CI.
For example, if any of the costs exceed the current costs by a defined margin (say 10%) then CI would fail.
^ This would also help us quickly catch changes in EVM functionality, detrimental changes to precompiles and weight changes.
^ Could be in another PR.

JasonTulp (approved):
Interesting, I kind of expected a bigger change. Nevertheless, good to have a solid baseline

PR #859, opened 2024-07-04
Update transaction-payment default

JasonTulp:
Sets the genesis config value for TransactionPaymentConfig fee_multiplier to equal what it evens out to on live chains. Previously this value was set far too high which will cause variance in the calculated fee values


zees-dev (approved):
Nice work; LGTM 👍

surangap (commented):
nice work.

PR #869, opened 2024-07-31
Stable rust toolchain migration

zees-dev:
## Context

Ideally we should migrate to stable rust for toolchain stability; this PR achieves that.

- There are discrepancies in the CI image used to produce the release and the rust version specified in the the `rust-toolchain.toml`
- The runtime wasm output/artifact uploaded for releases is currently the debug WASM (not release/optimized)
- Stable rust does not support the same `rustfmt.toml` as nightly; hence there's a bunch of formatting changes in this PR
  - Note all the changes to the `.rs` files are formatting changes due to the toolchain update
- The `xrpl-sdk-rust` main branch supports stable, hence the dep is updated it point to that branch; diff [here](https://github.com/futureversecom/xrpl-sdk-rust/pull/4/files)


.github/workflows/ci.yml
```diff
@@ -22,8 +22,11 @@ jobs:
     steps:
       - name: Check out
         uses: actions/checkout@v3
-      - name: Install toolchain
-        run: rustup show
```
JasonTulp, line 26: Why get rid of rustup show line?

JasonTulp (approved):
lgtm!

.github/workflows/ci.yml
zees-dev: This doesnt actually install the toolchain; just shows it. Using GH action instead to deterministically install the correct toolchain.

surangap (approved):
lgtm!

PR #873, opened 2024-08-15
Feat/non transferable flag

JasonTulp:
Adds additional flags to NFT and SFT pallet to allow collection level restrictions of transferring, minting and burning tokens within that collection

https://futureverse.atlassian.net/browse/TRN-467

### DIFF
--- a/pallet/nft/src/impls.rs
+++ b/pallet/nft/src/impls.rs
@@ -90,6 +90,10 @@ impl<T: Config> Pallet<T> {
 		new_owner: &T::AccountId,
 	) -> DispatchResult {
 		ensure!(current_owner != new_owner, Error::<T>::InvalidNewOwner);
+		ensure!(
+			<UtilityFlags<T>>::get(collection_id).transferable,
+			Error::<T>::TransferUtilityBlocked
+		);
 
 		CollectionInfo::<T>::try_mutate(collection_id, |maybe_collection_info| -> DispatchResult {
 			let collection_info =
@@ -479,6 +483,7 @@ impl<T: Config> Pallet<T> {
 			!<TokenLocks<T>>::contains_key((collection_id, serial_number)),
 			Error::<T>::TokenLocked
 		);
+		ensure!(<UtilityFlags<T>>::get(collection_id).burnable, Error::<T>::BurnUtilityBlocked);
 
 		CollectionInfo::<T>::try_mutate(collection_id, |maybe_collection_info| -> DispatchResult {
 			let collection_info =
--- a/pallet/nft/src/lib.rs
+++ b/pallet/nft/src/lib.rs
@@ -34,7 +34,9 @@ use frame_support::{
 	traits::{fungibles::Mutate, Get},
 	transactional, PalletId,
 };
-use seed_pallet_common::{OnNewAssetSubscriber, OnTransferSubscriber, Xls20MintRequest};
+use seed_pallet_common::{
+	utils::CollectionUtilityFlags, OnNewAssetSubscriber, OnTransferSubscriber, Xls20MintRequest,
+};
 use seed_primitives::{
 	AssetId, Balance, CollectionUuid, MetadataScheme, OriginChain, ParachainId, RoyaltiesSchedule,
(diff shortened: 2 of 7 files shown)


pallet/nft/src/lib.rs
```diff
@@ -153,6 +155,11 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type TokenLocks<T> = StorageMap<_, Twox64Concat, TokenId, TokenLockReason>;
 
+	/// Map from a collection to additional utility flags
+	#[pallet::storage]
+	pub type UtilityFlags<T> =
+		StorageMap<_, Twox64Concat, CollectionUuid, CollectionUtilityFlags, ValueQuery>;
```
zees-dev, line 161: Should this not be part of the `CollectionInformation` struct?
Otherwise there maybe too many structs associated to a single collection.
All the attributes of this struct apply to a collection; and in pretty much all cases you're going to be calling `CollectionInfo` storage item anyway.
In theory should should also reduce the reads and writes to storage locations - if this is part of the same struct.

pallet/nft/src/impls.rs
zees-dev, line 1: should also update the `do_mint` function too?

pallet/nft/src/lib.rs
```diff
@@ -631,6 +648,33 @@ pub mod pallet {
 			});
 			Ok(())
 		}
+
+		/// Set utility flags of a collection. This allows restricting certain operations on a
+		/// collection such as transfer, burn or mint
+		#[pallet::call_index(12)]
+		#[pallet::weight(T::WeightInfo::transfer())]
```
zees-dev, line 655: need benchmarks and relevant weight files.

pallet/sft/src/lib.rs
```diff
@@ -467,5 +482,32 @@ pub mod pallet {
 			Self::deposit_event(Event::<T>::MintPriceSet { token_id, payment_asset, mint_price });
 			Ok(())
 		}
+
+		/// Set utility flags of a collection. This allows restricting certain operations on a
+		/// collection such as transfer, burn or mint
+		#[pallet::call_index(12)]
+		#[pallet::weight(T::WeightInfo::set_mint_fee())]
```
zees-dev, line 489: need benchmarks and relevant weight files.

JasonTulp: https://github.com/futureversecom/trn-seed/pull/873/commits/348c2598c720db712efb63a5b5814b09207932c0

pallet/nft/src/lib.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/873/commits/348c2598c720db712efb63a5b5814b09207932c0

pallet/nft/src/impls.rs
JasonTulp: The flag shouldn't affect do_mint as this could disrupt bridged collections, we don't want to prevent a bridged token from coming through as it could cause the token to be lost in the process

pallet/nft/src/lib.rs
JasonTulp: I agree this is the best solution, and in an ideal world it would be included within CollectionInformation, but I feel we'd need to perform the huge migration of moving token ownership info into it's own struct. This piece of work was paused before the Substrate upgrade but I'm hoping to revisit this as with the Substrate upgrade there is better support for multi block migrations which will be safer. Unfortunately I think this piece of work is required before I'll have time to complete this migration, and I would prefer not to migrate CollectionInfo before then to include this field

zees-dev:
need to use GH workflow to generate weights

pallet/nft/src/lib.rs
JasonTulp: We could include migrating the UtilityFlags into CollectionInfo when we do the migration separating out token ownership however

surangap: maybe better to get an idea about the amount of data that needs to be migrated by looking at the mainnet?

surangap (commented):
lgtm. if moving new flags to the collectionInfo requires multiblock migration, can consider that later. if possible should do it with this PR

zees-dev (approved):
LGTM 👍

surangap (approved):

PR #877, opened 2024-09-04
TRN-521 Support setting payment delay per asset id in xrpl bridge

JCSanPedro:
Adds support for setting the payment threshold and delay per asset id (currently the same payment threshold and delay is set for all bridged assets).

**Changes**

- Update `PaymentDelay` storage value type to storage map type `AssetId -> (Balance, Block)`
- Update `set_payment_delay` to allow specifying `asset_id`
- Create migration for xrpl-bridge pallet storage
- Rerun benchmarks for xrpl-bridge


pallet/xrpl-bridge/src/weights.rs
```diff
@@ -18,8 +18,8 @@
 //! Autogenerated weights for pallet_xrpl_bridge
 //!
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
-//! DATE: 2024-09-03, STEPS: `50`, REPEAT: 20, LOW RANGE: `[]`, HIGH RANGE: `[]`
-//! HOSTNAME: `ip-172-31-102-147`, CPU: `Intel(R) Xeon(R) CPU E5-2686 v4 @ 2.30GHz`
+//! DATE: 2024-09-04, STEPS: `50`, REPEAT: 20, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! HOSTNAME: `Johns-Macbook-Pro.local`, CPU: `<UNKNOWN>`
```
JCSanPedro, line 22: Are we suppose to run benchmarks on CI because it depends on the values are dependent on the local machine it runs on?

runtime/src/migrations/xrpl_bridge.rs
```diff
@@ -45,187 +47,105 @@ impl OnRuntimeUpgrade for Upgrade {
...
+		weight = weight.saturating_add(<T as frame_system::Config>::DbWeight::get().reads(1));
 
-				Some(new_withdrawal)
-			},
-		);
+		if let Some(payment_delay) =
+			Value::unsafe_storage_get::<V3PaymentDelay<T>>(b"XRPLBridge", b"PaymentDelay")
+		{
+			weight = weight.saturating_add(<T as frame_system::Config>::DbWeight::get().writes(1));
+			PaymentDelay::<T>::insert(XrpAssetId::get(), payment_delay);
```
JasonTulp, line 103: Migration is really good although you're missing one part, because we are converting this storage item from a value to a map, the value will still be contained in storage and needs to be manually cleared. If you put the following in your test after the migration you'll see what I mean
```
assert_eq!(
	Value::unsafe_storage_get::<V3PaymentDelay<Runtime>>(
		b"XRPLBridge",
		b"PaymentDelay",
	),
	None
);
```
The above check should pass, but with your migration, that value still exists so it returns Some((111, 3344))

The way around this is to manually clear the storage value just before putting the new map in place with the following:
```
Value::unsafe_clear(b"XRPLBridge", b"PaymentDelay");
```
That way the old value is correctly removed

JasonTulp (commented):
Really nice work John, I'm very impressed with how quickly you're picking this up. I've left a few comments but once those are addressed this should be in a good state to merge!

runtime/src/migrations/xrpl_bridge.rs
```diff
@@ -29,11 +31,11 @@ impl OnRuntimeUpgrade for Upgrade {
 
 		let mut weight = <Runtime as frame_system::Config>::DbWeight::get().reads(2);
 
-		if onchain == 2 {
-			log::info!(target: "Migration", "XRPLBridge: Migrating from on-chain version 2 to on-chain version 3.");
-			weight += v3::migrate::<Runtime>();
+		if onchain < 4 {
```
JasonTulp, line 34: nit. the version here should be explicit i.e. == 3
This is to avoid any situations where the migration is run on previous versions when it shouldn't. (Although in this case there wouldn't be any consequences)

JasonTulp (approved):
lgtm!

runtime/src/migrations/xrpl_bridge.rs
JCSanPedro: Yeah I tried `onchain == 3` at first but it won't work when building the node from `main`, because the previous PR missed updating the onchain storage version to 3, and the onchain version is only set to 3 if the migration runs.

JasonTulp: Ah I see, good catch :)

pallet/xrpl-bridge/src/lib.rs
```diff
@@ -1155,7 +1164,7 @@ impl<T: Config> Pallet<T> {
 					Self::process_xrp_withdrawal(destination, amount, tx_fee, who.clone())?;
 
 				// Check if there is a payment delay and delay the payment if necessary
-				if let Some((payment_threshold, delay)) = PaymentDelay::<T>::get() {
+				if let Some((payment_threshold, delay)) = PaymentDelay::<T>::get(asset_id) {
```
surangap, line 1167: This is inside the XRPL scope. might need to take it out to apply for all tokens?

JCSanPedro: No intention was for it to apply to all tokens, I just missed it.

surangap (approved):
lgtm. nice work!
should run the CI bench again and should be good to merge.

PR #880, opened 2024-09-19
TRN-400 Emit event when listing is closed from accepted offer

JCSanPedro:
This PR modifies the `accept_offer` impl in the marketplace pallet so that it emits a `FixedPriceSaleClose` event in the case an offer is accepted, whilst a listing for that token already exists.

This allows front-end experiences to track this particular event when it occurs.

### DIFF
--- a/pallet/marketplace/src/impls.rs
+++ b/pallet/marketplace/src/impls.rs
@@ -385,7 +385,7 @@ impl<T: Config> Pallet<T> {
 							tokens: sale.tokens,
 							listing_id,
 							marketplace_id: sale.marketplace_id,
-							reason: FixedPriceClosureReason::VendorCancelled,
+							reason: FixedPriceClosureReason::OfferAccepted,
 						});
 					},
 					// Offer cannot be made on auctions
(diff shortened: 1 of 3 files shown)


pallet/marketplace/src/impls.rs
```diff
@@ -379,6 +379,19 @@ impl<T: Config> Pallet<T> {
 		if let Some(TokenLockReason::Listed(listing_id)) = T::NFTExt::get_token_lock(offer.token_id)
 		{
 			if let Some(listing) = <Listings<T>>::get(listing_id) {
+				match listing.clone() {
+					Listing::<T>::FixedPrice(sale) => {
+						Self::deposit_event(Event::<T>::FixedPriceSaleClose {
+							tokens: sale.tokens,
+							listing_id,
+							marketplace_id: sale.marketplace_id,
+							reason: FixedPriceClosureReason::VendorCancelled,
```
JasonTulp, line 388: We should add a separate closure reason for this case so we can easily track why it was closed

JasonTulp (approved):
lgtm! Nice job

JasonTulp:
Btw updating benchmarks for every PR isn't required, only if you add extra storage writes or reads for an extrinsic, otherwise the changes in benchmarks would be negligible

JCSanPedro:
> Btw updating benchmarks for every PR isn't required, only if you add extra storage writes or reads for an extrinsic, otherwise the changes in benchmarks would be negligible

Ah I run the benchmark as I thought emitting an extra event would contribute to the weight.

zees-dev (approved):

surangap (approved):

PR #881, opened 2024-09-25
Remove mapping from xrpl asset map

KarishmaBothara:
Extrinsic to remove mapping from assetIdToXRPL in xrplBridge pallet.

### DIFF
--- a/pallet/xrpl-bridge/src/lib.rs
+++ b/pallet/xrpl-bridge/src/lib.rs
@@ -17,7 +17,8 @@
 
 use crate::types::{
 	AssetWithdrawTransaction, DelayedPaymentId, DelayedWithdrawal, WithdrawTransaction,
-	XRPLCurrency, XrpTransaction, XrpWithdrawTransaction, XrplTicketSequenceParams, XrplTxData,
+	XRPLCurrency, XRPLCurrencyType, XrpTransaction, XrpWithdrawTransaction,
+	XrplTicketSequenceParams, XrplTxData,
 };
 use frame_support::{
 	fail,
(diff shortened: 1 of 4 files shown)


JasonTulp (commented):
Instead of adding a new extrinsic, why not change the existing extrinsic to have an option for xrpl_currency?

`xrpl_currency: Option<XRPLCurrency>`

We have use this pattern a few other times in the codebase

KarishmaBothara:
> Instead of adding a new extrinsic, why not change the existing extrinsic to have an option for xrpl_currency?
> 
> `xrpl_currency: Option<XRPLCurrency>`
> 
> We have use this pattern a few other times in the codebase

This would require migration?

pallet/xrpl-bridge/src/lib.rs
```diff
@@ -764,14 +768,27 @@ pub mod pallet {
...
+			xrpl_currency: Option<XRPLCurrency>,
 		) -> DispatchResult {
 			ensure_root(origin)?;
-			// Validate currency to prevent errors during withdrawal
-			ensure!(xrpl_currency.symbol.is_valid(), Error::<T>::InvalidCurrencyCode);
-			<AssetIdToXRPL<T>>::insert(asset_id, xrpl_currency);
-			<XRPLToAssetId<T>>::insert(xrpl_currency, asset_id);
-			Self::deposit_event(Event::XrplAssetMapSet { asset_id, xrpl_currency });
+			// Remove mapping if xrpl_currency is none
+			if xrpl_currency.is_none() {
```
surangap, line 775: nit - use the `let else` pattern to avoid the unwrap below.

surangap (commented):
lgtm. added a minor comment.

JasonTulp (approved):

surangap (approved):

PR #882, opened 2024-09-29
TRN-212 Exclude non-owner proxied FP transactions

JCSanPedro:
https://www.notion.so/futureverse/FRN-25-Privilege-system-non-functional-30ab4c8692b742659f43be471517d286

The privilege check within `Futurepas::proxy_extrinsic` are non-functional as they can be bypassed by calling `Proxy::proxy` using a delegate of the FP account.

In particular, a Futurepass account will have an `Owner` and also zero or more delegates. The owner and delegate accounts can perform extrinsics on behalf of the futurepass account via proxy extrinsics. There are two ways to do this:

1. Invoking the `Futurepass::proxy_extrinsic` call
2. Or directly calling `Proxy::proxy`

The `Futurepass::proxy_extrinsic` call has a guard that prevents certain extrinsics from being executed unless the proxy caller is the owner of the Futurepass. These in particular are
* `Futurepass::register_delegate_with_signature`
* `Futurepass::unregister_delegate`
* `Futurepass::transfer_futurepass`
Effectively, we only want the owner to be able to invoke the above extrinsics (a delegate can call `unregister_delegate` but should call directly, not proxied). However a delegate account can bypass this protection by calling `Proxy::proxy` directly.

This allows for an unwanted case such as a delegate account being able to invoke `transfer_futurepass`.

To prevent this, we update the `InstanceFilter` impl to only allow the `Owner` `ProxyType` to make certain proxy calls. The `filter` implementation will not execute any proxy transactions if they are one of the listed extrinsics, and the `ProxyType` is not of the [Owner](https://github.com/futureversecom/trn-seed/blob/main/runtime/src/impls.rs#L702) variant.

The filter first checks if the `proxy` call is calling another `proxy` extrinsic or a `futurepass` extrinsic. In this case only the `register_delegate_with_signature`, `unregister_delegate`, and `transfer_futurepass` extrinsics are allowed. We the additionally ensure that the `ProxyType` is of `Owner`

```
if matches!(c, RuntimeCall::Proxy(..) | RuntimeCall::Futurepass(..)) {
		// Whitelist currently includes
		// pallet_futurepass::Call::register_delegate_with_signature,
		// pallet_futurepass::Call::unregister_delegate
		if !matches!(
			c,
			RuntimeCall::Futurepass(
				pallet_futurepass::Call::register_delegate_with_signature { .. }
			) | RuntimeCall::Futurepass(pallet_futurepass::Call::unregister_delegate { .. })
				| RuntimeCall::Futurepass(pallet_futurepass::Call::transfer_futurepass { .. })
		) {
			return false;
		}

		return self == &ProxyType::Owner;
}
```

### DIFF
--- a/pallet/futurepass/src/tests.rs
+++ b/pallet/futurepass/src/tests.rs
@@ -1632,6 +1632,80 @@ fn proxy_extrinsic_to_proxy_pallet_fails() {
 }
 
 #[test]
+fn cannot_bypass_proxy_extrinsic_via_proxy() {
+	let funder = create_account(1);
+	let endowed = [(funder, 1_000_000)];
+
+	TestExt::<Test>::default()
+		.with_balances(&endowed)
+		.with_xrp_balances(&endowed)
+		.build()
+		.execute_with(|| {
+			let owner = create_account(2);
+			let (signer, delegate) = create_random_pair();
+
+			let proxy_type = ProxyType::Any;
+			let deadline = 200;
+
+			// fund owner
+			transfer_funds(
+				MOCK_NATIVE_ASSET_ID,
+				&funder,
+				&owner,
+				FP_CREATION_RESERVE + FP_DELEGATE_RESERVE + 1,
+			);
+
+			// create FP for owner
+			assert_ok!(Futurepass::create(RuntimeOrigin::signed(owner), owner));
+			let futurepass = Holders::<Test>::get(&owner).unwrap();
+
+			let signature = signer
+				.sign_prehashed(
+					&Futurepass::generate_add_delegate_eth_signed_message(
+						&futurepass,
+						&delegate,
+						&proxy_type,
+						&deadline,
+					)
+					.unwrap()
+					.1,
+				)
+				.0;
+
+			// register delegate
+			assert_ok!(Futurepass::register_delegate_with_signature(
+				RuntimeOrigin::signed(owner),
+				futurepass,
+				delegate,
+				proxy_type,
+				deadline,
+				signature,
+			));
+
+			// fund futurepass
+			transfer_funds(MOCK_NATIVE_ASSET_ID, &funder, &futurepass, 1000);
+
+			let other = create_account(4);
+			let inner_call = Box::new(MockCall::Futurepass(Call::transfer_futurepass {
+				current_owner: owner,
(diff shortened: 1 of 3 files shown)


zees-dev (approved):
LGTM 👍

runtime/src/impls.rs
```diff
@@ -786,6 +786,10 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
 			) {
 				return false;
 			}
+
+			// the whitelisted calls above should only be able to be called by
+			// the owner of the futurepass
+			return self == &ProxyType::Owner;
```
surangap, line 792: So the functionality requirement overall is that only the owner should be able to proxy the whitelist either via `futurepass::proxy_extrinsic()` or `proxy::proxy()`?
Do you think this change would make some of the checks in `futurepass::proxy_extrinsic()` redundant?

pallet/futurepass/src/tests.rs
```diff
@@ -1631,6 +1631,80 @@ fn proxy_extrinsic_to_proxy_pallet_fails() {
...
+				new_owner: Some(other),
+			}));
+			assert_ok!(pallet_proxy::Pallet::<Test>::proxy(
+				RuntimeOrigin::signed(delegate),
+				futurepass,
+				None,
+				inner_call
+			));
+
+			// the delegate tried to transfer the futurepass, but because it was
```
surangap, line 1700: Maybe add checks to show that the owner still can do it with Proxy::proxy()

surangap (commented):
lgtm. added a minor comment

runtime/src/impls.rs
JCSanPedro: It's somewhat redundant as the filter would prevent a non-owner proxy call from going through, but I think it's still valuable to retain the checks in `proxy_extrinsic` as it will return a proper error (`NotFuturepassOwner`) if the call is made from a non-owner account, whereas calling `Proxy::proxy` just silently no-ops. And I think `proxy_extrinsic` would be the preferred way to make proxied FP calls anyway. 

The filter is just to prevent the particular case of an account sidestepping `proxy_extrinsic` by calling `Proxy::proxy` directly.

surangap (approved):
nice work👏

PR #886, opened 2024-10-03
🦆 Multi-Block Migration

JasonTulp:
Adds migration pallet which handles multi-block migrations for other pallets
Adds multi-block migration for XLS20 token ID from incorrectly stored 64 byte array to 32 byte array

More details can be found on the Migration pallet here:
https://www.notion.so/futureverse/Migration-Pallet-1260cb4dab3d80a995a2c5856a4c4afa


pallet/xls20/Cargo.toml
```diff
@@ -19,6 +19,7 @@ scale-info = { workspace = true }
 pallet-nft = { workspace = true }
 seed-primitives = { workspace = true }
 seed-pallet-common = { workspace = true }
+log = "0.4.22"
```
zees-dev, line 22: use from workspace:
```suggestion
log = { workspace = true }
```

runtime/src/migrations/xls20_multi.rs
```diff
@@ -0,0 +1,229 @@
...
+				Ok(new_value) => {
+					pallet_xls20::Xls20TokenMap::<T>::insert(key1, key2, new_value);
+				},
+				Err(e) => {
+					// If we encounter an error during the conversion, we must insert some
+					// fake value, we can't remove the token as it will cause unexpected results
+					// with the iter process
+					log::error!(target: LOG_TARGET, "🦆 Error migrating token_id ({:?},{:?}) : {:?}", key1, key2, e);
+					let fake_value = [0u8; 32];
+					pallet_xls20::Xls20TokenMap::<T>::insert(key1, key2, fake_value);
```
zees-dev, line 89: maybe call this `default value` instead of `fake value`

pallet/migration/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+
+	/// The delay between migration blocks
+	#[pallet::storage]
+	pub type BlockDelay<T> = StorageValue<_, u32, OptionQuery>;
+
+	/// Default value is 100 which is on the conservative side
+	#[pallet::type_value]
+	pub fn DefaultBlockLimit() -> u32 {
+		100
+	}
```
zees-dev, line 104: nit: could make this configurable - so that runtime upgrades will not be required to set this value.

pallet/migration/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+			Ok(())
+		}
+
+		/// Set the block limit for multi-block migrations. The block limit is the maximum number
+		/// of individual items to migrate in a single block. Will still respect maximum weight rules.
+		#[pallet::call_index(2)]
+		#[pallet::weight(T::WeightInfo::enable_migration())]
+		pub fn set_block_limit(origin: OriginFor<T>, block_limit: u32) -> DispatchResult {
+			ensure_root(origin)?;
+			BlockLimit::<T>::put(block_limit);
```
zees-dev, line 207: nit: also ensure this is `> 0`?

pallet/migration/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+				false => Self::deposit_event(Event::MigrationDisabled),
+			}
+			Ok(())
+		}
+
+		/// Set the block delay for multi-block migrations. The block delay is the number
+		/// of blocks between each migration step. If set to None, the migration will run every block.
+		/// The delay must be greater than 1.
+		#[pallet::call_index(1)]
+		#[pallet::weight(T::WeightInfo::enable_migration())]
```
zees-dev, line 187: weight needs updating

pallet/migration/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+				None => BlockDelay::<T>::kill(),
+			}
+			Self::deposit_event(Event::BlockDelaySet { block_delay });
+			Ok(())
+		}
+
+		/// Set the block limit for multi-block migrations. The block limit is the maximum number
+		/// of individual items to migrate in a single block. Will still respect maximum weight rules.
+		#[pallet::call_index(2)]
+		#[pallet::weight(T::WeightInfo::enable_migration())]
```
zees-dev, line 204: weight needs updating

pallet/migration/src/weights.rs
zees-dev, line 1: needs to be generated on canonical machine

runtime/src/lib.rs
```diff
@@ -1364,6 +1365,20 @@ impl pallet_crowdsale::Config for Runtime {
...
+parameter_types! {
+	// The upper limit of weight used per block for migrations is 10%
+	// Note, this could still be smaller if we set a smaller BlockLimit within pallet-migration
+	pub MaxMigrationWeight: Weight = Perbill::from_percent(10) * MAXIMUM_BLOCK_WEIGHT;
+}
+
+impl pallet_migration::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	// Set to NoopMigration if no migration is in progress
+	type CurrentMigration = migrations::xls20_multi::Xls20Migration<Runtime>;
```
zees-dev, line 1377: Instead of this, there should be another struct in the migrations module - which wraps all the migrations that need to be performed - since there maybe more than one multi-block migration.
In the implementation of that is where you'd need to call the migrations of the specific pallets (xls20 in this case).
Otherwise multi-block migrations will only be applicable to a single pallet.

Cargo.toml
```diff
@@ -110,6 +110,7 @@ pallet-assets-ext-rpc = { path = "pallet/assets-ext/rpc", default-features = fal
 pallet-assets-ext-rpc-runtime-api = { path = "pallet/assets-ext/rpc/runtime-api", default-features = false }
 seed-pallet-common = { path = "pallet/common", default-features = false }
 pallet-crowdsale = { path = "pallet/crowdsale", default-features = false }
+pallet-migration = { path = "pallet/migration", default-features = false }
```
zees-dev, line 113: lets try retain alphabetical ordering here

runtime/src/lib.rs
surangap: Another approach could be to handle multiple pallet migrations in the passed migration struct. that way we could write the migration as required and the migration pallet just needs to pass the control to that migration struct via the`MigrationStep` trait.

pallet/migration/src/lib.rs
JasonTulp: This is just the default value, but it is configurable using set_block_limit extrinsic

runtime/src/lib.rs
JasonTulp: As we discussed at the start of this work, I have only focussed on single pallet functionality for now. If we have a need for supporting multiple pallets simultaneously (Which won't be in the foreseeable future) we can expand the functionality. For now this is wasted effort as we only need one pallet at a time

primitives/src/migration.rs
```diff
@@ -0,0 +1,95 @@
...
+	fn version_check() -> bool;
+
+	/// Called when the migration is complete.
+	fn on_complete();
+
+	/// Returns the maximum weight that can be consumed in a single step.
+	fn max_step_weight() -> Weight;
+
+	/// Convert from the OldStorageValue to the NewStorageValue
+	fn convert(old: Self::OldStorageValue) -> Result<Self::NewStorageValue, &'static str>;
```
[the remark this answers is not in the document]
JasonTulp: Separating out convert made it easier to test as well as making it clearer which logic performs the actual conversion. This may not be a one size fits all solution though as you say as some storage items in future may need multiple conversions

Cargo.toml
JasonTulp: https://github.com/futureversecom/trn-seed/pull/886/commits/c8ede1291e2f5f901710bb093e04a481c7194125

pallet/migration/src/lib.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/886/commits/c8ede1291e2f5f901710bb093e04a481c7194125

pallet/migration/src/lib.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/886/commits/c8ede1291e2f5f901710bb093e04a481c7194125

pallet/migration/src/lib.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/886/commits/c8ede1291e2f5f901710bb093e04a481c7194125

runtime/src/migrations/xls20_multi.rs
JasonTulp: https://github.com/futureversecom/trn-seed/pull/886/commits/c8ede1291e2f5f901710bb093e04a481c7194125

pallet/xls20/Cargo.toml
JasonTulp: https://github.com/futureversecom/trn-seed/pull/886/commits/c8ede1291e2f5f901710bb093e04a481c7194125

surangap (approved):
lgtm. Nice work👍. Let's get the CI tests working.

PR #889, opened 2024-10-08
EIP-2470 support + Universal deployer support

zees-dev:
## Context

This PR adds support for [EIP-2470](https://eips.ethereum.org/EIPS/eip-2470) and [Universal deployer](https://gist.github.com/Agusx1211/de05dabf918d448d315aa018e2572031) via a runtime upgrade.
Bytcode would be manually injected into the EVM - to mimic the effects of a transaction (post contract deployment).

### Why can we not simply create a transaction which deploys these contracts?
Contract deployment (using [Nick's method](https://weka.medium.com/how-to-send-ether-to-11-440-people-187e332566b7)) requires that the canonical transaction be constructed with `100 gwei` `gasPrice`.
This is not possible on TRN EVM due to the base gas price being `7500 gwei`; hence the transaction for deployment of the contract(s) cannot be executed.

## Changelog

- Create EIP-2470 factory at `0xce0042B868300000d44A59004Da54A005ffdcf9f` with relevant bytecode
- Increment nonce of EIP-2470 contract deployer `0xBb6e024b9cFFACB947A71991E386681B1Cd1477D`
- Create Universal deployer factory at `0x1b926fbb24a9f78dcdd3272f2d86f5d0660e59c0` with relevant bytecode
- Increment nonce of Universal deployer EOA account (which deploys the contract) `0x9c5a87452d4FAC0cbd53BDCA580b20A45526B3AB`
- Migration tests

Note: The migrations will always be run if the factory contract addresses are empty (i.e. they don't have the relevant bytecode - implying that the factory is not deployed on the chain).


runtime/src/migrations/evm.rs
```diff
@@ -0,0 +1,233 @@
...
+
+const UNIVERSAL_DEPLOYER_EOA_ADDRESS: &str = "9c5a87452d4FAC0cbd53BDCA580b20A45526B3AB";
+const UNIVERSAL_DEPLOYER_CONTRACT_ADDRESS: &str = "1b926fbb24a9f78dcdd3272f2d86f5d0660e59c0";
+const UNIVERSAL_DEPLOYER_CONTRACT_DATA: &str = "6080604052348015600f57600080fd5b50609980601d6000396000f3fe60a06020601f369081018290049091028201604052608081815260009260609284918190838280828437600092018290525084519495509392505060208401905034f5604080516001600160a01b0383168152905191935081900360200190a0505000fea26469706673582212205a310755225e3c740b2f013fb6343f4c205e7141fcdf15947f5f0e0e818727fb64736f6c634300060a0033";
```
surangap, line 27: why do we require this? is it require for `EIP2470`? or is it something similar?

surangap: alright i think the answer is in the PR description
`This PR adds support for [EIP-2470](https://eips.ethereum.org/EIPS/eip-2470) and [Universal deployer](https://gist.github.com/Agusx1211/de05dabf918d448d315aa018e2572031) via a runtime upgrade.`

surangap (commented):
lgtm. Had a question here -> https://github.com/futureversecom/trn-seed/pull/889/files#r1791160879

runtime/src/migrations/evm.rs
zees-dev: You need to execute the transaction; then analyze the bytecode of the deployed contract.
I did this on a local EVM instance.

The Universal deployer is a variation of the EIP-2470 factory - related to [EIP-1820](https://eips.ethereum.org/EIPS/eip-1820) i believe (although im just following the gist requirements).

runtime/src/migrations/evm.rs
```diff
@@ -0,0 +1,233 @@
...
+	where
+		<Runtime as frame_system::Config>::AccountId: From<H160>,
+	{
+		let mut weight = <Runtime as frame_system::Config>::DbWeight::get().reads(6);
+
+		// r: 3 + 3, w: 0
+		if !(EVM::is_account_empty(&deployer_eoa) && EVM::is_account_empty(&contract_address)) {
+			log::info!(target: "Migration", "No migration was done, however migration code needs to be removed.");
+			return weight;
+		}
```
JasonTulp, line 157: This should probably be moved higher up in the migration process, looks like if the migration attempts to execute a second time, it will count as 12 reads (deploy_contract gets called with noop twice from migrate)
Might be clearer to move the repeat check to `on_runtime_upgrade()`

runtime/src/migrations/evm.rs
```diff
@@ -0,0 +1,233 @@
...
+
+		// Set nonce for deployer EOA
+		// r: 0, w: 1
+		frame_system::Pallet::<Runtime>::inc_account_nonce(
+			&<Runtime as pallet_evm::Config>::AddressMapping::into_account_id(deployer_eoa),
+		);
+
+		// Set data (bytecode) and nonce for contract
+		EVM::create_account(contract_address, hex::decode(contract_bytecode).unwrap()); // r: 1, w: 2
+																				// r: 0, w: 1
```
JasonTulp, line 167: Might be my GH but this comment is very far to the right

JasonTulp (approved):
Looks good! Nice work, hard to verify contract_data but execution looks correct, verification will be done through testing on Sprout

runtime/src/migrations/evm.rs
zees-dev: probably gh since it appears fine, also it is cargo formatted

runtime/src/migrations/evm.rs
zees-dev: addressed: https://github.com/futureversecom/trn-seed/pull/889/commits/1dca76c2d137ba3525922ee25e1ad8041ffdd21a

PR #893, opened 2024-10-10
TRN-609 Improve `transact` error handling

JCSanPedro:
Front-end developers often encountered a misleading error message `1010: Invalid Transaction: Transaction has a bad signature` when working the `xrpl` pallet. This would usually be due to an unrelated problem, such as the sending account not having sufficient balance to pay for the extrinsic. Seeing the incorrect error message could be very confusing.  

The above error message is correlated with the error `InvalidTransaction::BadProof`. The runtime would mistakingly return this error because the `xrpl` (and `doughnut`) pallet have additional extrinsic validation applied to their runtime calls, implemented via the [SelfContainedCall](https://github.com/polkadot-evm/frontier/blob/master/primitives/self-contained/src/lib.rs#L36) trait that was returning the wrong value. The method `validate_self_contained` returns a type `Option<TransactionValidityError>`, where if the return value was `None`, the `validate` implementation for CheckedExtrinsics from evm frontier would default to `InvalidTransaction::BadProof` https://github.com/polkadot-evm/frontier/blob/master/primitives/self-contained/src/checked_extrinsic.rs#L93-L95.

This PR aims to ensure the correct `InvalidTransaction` variant is propagated when the runtime call is validated.

#### Changed:
doughtnut: Invalid transactions should be correct variant instead of catch-all `BadProof`
xrpl: Invalid transactions should be correct variant instead of catch-all `BadProof`

### DIFF
--- a/pallet/xrpl/src/lib.rs
+++ b/pallet/xrpl/src/lib.rs
@@ -143,57 +143,60 @@ where
 		len: usize,
 	) -> Option<TransactionValidity> {
 		if let Call::transact { encoded_msg, signature, call } = self {
-			let (nonce, tip) = validate_params::<T>(encoded_msg.as_bytes_ref(), signature.as_bytes_ref(), call)
-				.map_err(|e| {
-					log!(info, "⛔️ validate_self_contained: failed to validate params: {:?}", e);
-					InvalidTransaction::Call
-				})
-				.ok()?;
+			let validate = || -> TransactionValidity {
+				let (nonce, tip) = validate_params::<T>(encoded_msg.as_bytes_ref(), signature.as_bytes_ref(), call)
+					.map_err(|e| {
+						log!(info, "⛔️ validate_self_contained: failed to validate params: {:?}", e);
+						InvalidTransaction::Call
+					})?;
 
-			let validations: XRPLValidations<T> = (
-				CheckNonZeroSender::new(),
-				CheckSpecVersion::<T>::new(),
-				CheckTxVersion::<T>::new(),
-				CheckEra::<T>::from(Era::immortal()),
-				CheckWeight::new(),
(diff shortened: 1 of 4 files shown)


zees-dev (approved):
Nice; LGTM 👍

pallet/xrpl/src/lib.rs
```diff
@@ -143,57 +143,60 @@ where
...
-				if eoa == H160::zero() {
-					log!(info, "⛔️ failed to get EOA from futurepass address");
-					return None;
+				// validate signed extensions using EOA - for futurepass based transactions
+				if <T as pallet::Config>::FuturepassLookup::check_extrinsic(&call, &()) {
+					// this implies that the origin is futurepass address; we need to get the EOA associated with it
+					let eoa = <T as pallet::Config>::FuturepassLookup::unlookup(*origin);
+					if eoa == H160::zero() {
+						log!(info, "⛔️ failed to get EOA from futurepass address");
+						// return None;
```
JasonTulp, line 173: Why comment out the return?

JCSanPedro: Nice catch - did it to test some stuff but mistakenly left the comment in.

JasonTulp: Sweet I had a feeling that was the case :slightly_smiling_face:

JCSanPedro:
> Although interesting that only one test needs changing in doughnuts, is this an indication that we need better test coverage

Yes I think so. I found that for the Doughnut pallet tests, the tx would fail at `check_self_contained` before failing at `validate_self_contained`. Needs more exploration to figure out how to correctly write the tests so that it hits the code we expect it to.

JasonTulp (approved):

PR #894, opened 2024-10-16
EIP2470 fix

zees-dev:
## Changelog

Using factory address to check empty contract - instead of an EOA since it may have funds (is not empty)


JasonTulp (approved):
lgtm! Will there be issues with the other account being already migrated?

zees-dev:
> lgtm! Will there be issues with the other account being already migrated?

Should not be as other accounts also have not migrated.

PR #897, opened 2024-10-21
XRPL -> TRN NFT bridging

surangap:
Adds support for NFT bridging from XRPL to TRN.

- This can be achieved by calling submit_transaction in pallet_xrpl_bridge
- This then gets routed to pallet_xls20 using the Xls20Ext trait (T::Xls20Ext::deposit_xls20_token)
- In pallet-xls20 if the token exists, it will transfer, otherwise it will mint the token or create a new collection if it's the first time a token is bridged. 
- For all existing xls20 compatible tokens, we need to call `set_collection_mappings` to initiate the mappings of those tokens to collections on TRN


pallet/xls20/src/lib.rs
```diff
@@ -321,3 +357,89 @@ impl<T: Config> Xls20MintRequest for Pallet<T> {
...
+			// There is already a collection mapping for this token, this means it is either
+			// a TRN originated token that has been bridged over, or it is a token form XRPL
+			// from a collection that has been bridged before.
+
+			if Xls20TokenMap::<T>::get(collection_uuid, xls20_token.sequence)
+				== Some(xls20_token_id)
+			{
+				// The token ID exists in an existing mapping. Either this token originated on TRN
+				// or it has been bridged in the past. The pallet should own the token
+				T::NFTExt::do_transfer(
```
surangap, line 390: Is this correct? at any point the pallet does not own a token?
Further, we burn the nft when bridged from TRN -> XRPL here -> https://github.com/futureversecom/trn-seed/pull/898/files#r1823513401
also we don't remove from `Xls20TokenMap` when it burns. probably should add that as well.

pallet/xrpl-bridge/src/lib.rs
```diff
@@ -868,96 +863,59 @@ impl<T: Config> Pallet<T> {
...
 						ledger_index,
 						transaction_hash.clone(),
 					));
+					weight
 				},
-				_ => {
-					Self::deposit_event(Event::NotSupportedTransaction);
-					continue;
+				Err((weight, e)) => {
+					Self::deposit_event(Event::ProcessingFailed(
```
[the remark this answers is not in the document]
JasonTulp: https://github.com/futureversecom/trn-seed/pull/897/commits/23d4c0792ec381569862c5028bb70aaaaadda34a

pallet/xls20/src/lib.rs
JasonTulp: Swapped to do_mint instead

JasonTulp: https://github.com/futureversecom/trn-seed/pull/897/commits/c4169028d8f49efe63e077b9689b815e7f3ecda3

PR #901, opened 2024-11-03
XRPL -> TRN NFT bridging

JasonTulp:
Adds support for NFT bridging from XRPL to TRN.

This can be achieved by calling submit_transaction in pallet_xrpl_bridge
This then gets routed to pallet_xls20 using the Xls20Ext trait (T::Xls20Ext::deposit_xls20_token)
In pallet-xls20 if the token exists, it will transfer, otherwise it will mint the token or create a new collection if it's the first time a token is bridged.
For all existing xls20 compatible tokens, we need to call set_collection_mappings to initiate the mappings of those tokens to collections on TRN


surangap (approved):
lgtm

PR #917, opened 2025-01-08
TRN-808 Sylo Data Pallet

JCSanPedro:
# Description

The PR implements and adds the Sylo Data Verification Pallet to the runtime.

Protocol RFC - https://www.notion.so/futureverse/RFC-Sylo-Data-Verification-Protocol-14a0cb4dab3d808389eddaab37a2c767

The pallet allows users to register and manage **resolvers**, and also create and manage **validation records**.

## Resolvers

A resolver is intended to be an external service capable of storing and providing off-chain data. The pallet allows a user to register a resolver, which would include an identifier and a list of service endpoints. A resolver is identified by a `ResolverId`, which loosely follows the [DID Specification](https://www.w3.org/TR/did-core/).

The `ResolverId` is comprised of two components: a `method` and `identifier`. The `method` describes what kind of resolver it is. All resolvers registered on chain are considered "sylo" resolvers, and will all use the same reserved string for their resolver method.

The account which registers the resolver is considered the controller of the register, and is capable of modifying or removing the resolver later.

## Validation Records

The pallet allows a user to create a validation record through the `create_validation_record` extrinsic. A validation record references an externally stored piece of data, and is intended to be used to validate the integrity and authenticity of offchain data.

A validation record will have the following attributes:

* `data_id` - A string representing the id. Validation records are stored in a double map, using the `data_id` as well as the author's account id.
* `author` - The account which created the record. Will have the ability to modify and or delete the record in the future.
* `resolvers` - A list of `ResolverIds` which reference how to fetch the data. All resolvers using the reserved sylo method must be registered onchain.
* `data_type` - A string acting as a hint to clients as to what type of data it is.
* `tags` - A list of strings acting as additional metadata information
* `entries` A list of validation record entry. Each entry represents a distinct version of the data. It includes a checksum for integrity, and also a block number for when the entry was created.

## Notes

- **Payment** - All extrinsics related to managing validation records and resolvers are paid using Sylo tokens. This is achieved by modifying the `OnChargeTransaction` implementation for the FeeProxy, to detect sylo related extrinsics, and force the swap.

---

# Release Notes

## Key Changes

Adds the Sylo Data Verification Pallet to the runtime

## Type of Change

- [x] Runtime Changes

#### Added

- **SyloDataVerification: register_resolver**
- **SyloDataVerification: update_resolver**
- **SyloDataVerification: deregister_resolver**
- **SyloDataVerification: create_validation_record**
- **SyloDataVerification: add_validation_record_entry**
- **SyloDataVerification: update_validation_record**
- **SyloDataVerification: delete_validation_record**


runtime/src/lib.rs
```diff
@@ -596,6 +596,25 @@ impl pallet_preimage::Config for Runtime {
 	type ByteDeposit = PreimageByteDeposit;
 }
 
+parameter_types! {
+	pub const MaxResolvers: u32 = 10;
+	pub const MaxTags: u32 = 10;
+	pub const MaxEntries: u32 = 100;
+	pub const MaxServiceEndpoints: u32 = 10;
+	pub const SyloStringLimit: u32 = 500;
```
zees-dev, line 604: could reduce space for these vars:
```suggestion
	pub const MaxResolvers: u8 = 10;
	pub const MaxTags: u8 = 10;
	pub const MaxEntries: u8 = 100;
	pub const MaxServiceEndpoints: u8 = 10;
	pub const SyloStringLimit: u16 = 500;
```

runtime/src/weights/pallet_sylo.rs
zees-dev, line 1: weights need to be generated via CI - the runner would commit the updated weights

pallet/common/src/test_utils.rs
```diff
@@ -723,3 +723,26 @@ macro_rules! impl_pallet_scheduler_config {
 		}
 	};
 }
+
+#[macro_export]
+macro_rules! impl_pallet_sylo_config {
```
zees-dev, line 728: nice; thanks for this

pallet/fee-proxy/Cargo.toml
```diff
@@ -23,6 +23,7 @@ pallet-evm = { workspace = true }
 pallet-assets-ext = { workspace = true }
 pallet-dex = { workspace = true }
 pallet-futurepass = { workspace = true }
+pallet-sylo = { workspace = true }
```
zees-dev, line 26: may need an entry in the `std = [` feature below.

pallet/fee-proxy/src/impls.rs
zees-dev, line 1: will wait for validated E2E tests to approve changes here

pallet/sylo/src/types.rs
zees-dev, line 1: I dont think we need explicit/restrictive trait-bounds for all the types.
Can take a look at the `pub struct SaleInformation<AccountId, BlockNumber> {` in `pallet/crowdsale/src/types.rs` file to see struct declarations which just consume a generic.

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+		},
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::call_index(0)]
+		#[pallet::weight({
+			T::WeightInfo::set_payment_asset()
+		})]
+		pub fn set_payment_asset(origin: OriginFor<T>, payment_asset: AssetId) -> DispatchResult {
```
zees-dev, line 193: event needs to be emitted for all state changes.

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+			ensure_root(origin)?;
+			<SyloAssetId<T>>::put(payment_asset);
+			Ok(())
+		}
+
+		#[pallet::call_index(1)]
+		#[pallet::weight({
+			T::WeightInfo::set_sylo_resolver_method()
+		})]
+		pub fn set_sylo_resolver_method(
```
zees-dev, line 203: same here (emit event)

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::call_index(0)]
+		#[pallet::weight({
+			T::WeightInfo::set_payment_asset()
+		})]
+		pub fn set_payment_asset(origin: OriginFor<T>, payment_asset: AssetId) -> DispatchResult {
+			ensure_root(origin)?;
```
zees-dev, line 194: rather than using root; use the `ApproveOrigin` account (passed in via runtime)

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+
+		#[pallet::call_index(1)]
+		#[pallet::weight({
+			T::WeightInfo::set_sylo_resolver_method()
+		})]
+		pub fn set_sylo_resolver_method(
+			origin: OriginFor<T>,
+			resolver_method: BoundedVec<u8, T::StringLimit>,
+		) -> DispatchResult {
+			ensure_root(origin)?;
```
zees-dev, line 207: same here (use approved origin)

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+			origin: OriginFor<T>,
+			identifier: BoundedVec<u8, T::StringLimit>,
+			service_endpoints: BoundedVec<ServiceEndpoint<T::StringLimit>, T::MaxServiceEndpoints>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			ensure!(
+				<Resolvers<T>>::get(identifier.clone()).is_none(),
+				Error::<T>::ResolverAlreadyRegistered
+			);
```
zees-dev, line 226: nit:

```suggestion
			<Resolvers<T>>::get(&identifier).ok_or(()).map_err(|_| Error::<T>::ResolverAlreadyRegistered)?;
```

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+
+			<Resolvers<T>>::insert(identifier.clone(), resolver);
+
+			Self::deposit_event(Event::ResolverUpdated {
+				id: identifier.to_vec(),
+				controller: who,
+				service_endpoints,
+			});
+
+			Ok(())
```
zees-dev, line 268: nit: instead of getting and inserting:

```suggestion
			<Resolvers<T>>::try_mutate(&identifier, |maybe_resolver| {
                let resolver = maybe_resolver.as_mut().ok_or(Error::<T>::ResolverNotRegistered)?;
                ensure!(who == resolver.controller, Error::<T>::NotController);
                resolver.service_endpoints = service_endpoints.clone();
                
                Self::deposit_event(Event::ResolverUpdated {
                    id: identifier.into_vec(),
                    controller: who,
                    service_endpoints,
                });
            Ok(())
        })
```

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+			identifier: BoundedVec<u8, T::StringLimit>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			let resolver =
+				<Resolvers<T>>::get(identifier.clone()).ok_or(Error::<T>::ResolverNotRegistered)?;
+
+			ensure!(who == resolver.controller, Error::<T>::NotController);
+
+			<Resolvers<T>>::remove(identifier.clone());
```
zees-dev, line 286: nit: unnecessary clone (memory allocations) can be removed i think
```suggestion
			let resolver = <Resolvers<T>>::get(&identifier).ok_or(Error::<T>::ResolverNotRegistered)?;
            ensure!(who == resolver.controller, Error::<T>::NotController);
            
            <Resolvers<T>>::remove(&identifier);
```

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+			data_id: BoundedVec<u8, T::StringLimit>,
+			resolvers: BoundedVec<ResolverId<T::StringLimit>, T::MaxResolvers>,
+			data_type: BoundedVec<u8, T::StringLimit>,
+			tags: BoundedVec<BoundedVec<u8, T::StringLimit>, T::MaxTags>,
+			checksum: H256,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			ensure!(
+				<ValidationRecords<T>>::get(who.clone(), data_id.clone()).is_none(),
```
zees-dev, line 308:
```suggestion
				<ValidationRecords<T>>::get(&who, &data_id).is_none(),
```

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+				if resolver.method == reserved_method {
+					ensure!(
+						<Resolvers<T>>::get(resolver.identifier).is_some(),
+						Error::<T>::ResolverNotRegistered
+					);
+				}
+			}
+
+			Ok(())
+		}
```
zees-dev, line 453: nit:
```suggestion
		pub fn validate_sylo_resolvers(
            resolvers: &BoundedVec<ResolverId<T::StringLimit>, T::MaxResolvers>,
        ) -> DispatchResult {
            let reserved_method = <SyloResolverMethod<T>>::get();

            resolvers
                .iter()
                .filter(|resolver| resolver.method == reserved_method)
                .try_for_each(|resolver| {
                    ensure!(
                        <Resolvers<T>>::contains_key(&resolver.identifier),
                        Error::<T>::ResolverNotRegistered
                    );
                    Ok(())
                })
        }
```

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+			checksum: H256,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			ensure!(
+				<ValidationRecords<T>>::get(who.clone(), data_id.clone()).is_none(),
+				Error::<T>::RecordAlreadyCreated
+			);
+
+			Self::validate_sylo_resolvers(resolvers.clone())?;
```
zees-dev, line 312:
```suggestion
			Self::validate_sylo_resolvers(&resolvers)?;
```

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+				resolvers,
+				data_type,
+				tags,
+				entries: BoundedVec::truncate_from(vec![ValidationEntry {
+					checksum,
+					block: current_block,
+				}]),
+			};
+
+			<ValidationRecords<T>>::insert(who.clone(), data_id.clone(), record);
```
zees-dev, line 327:
```suggestion
			<ValidationRecords<T>>::insert(&who, &data_id, record);
```

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+			T::WeightInfo::add_validation_record_entry()
+		})]
+		pub fn add_validation_record_entry(
+			origin: OriginFor<T>,
+			data_id: BoundedVec<u8, T::StringLimit>,
+			checksum: H256,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			let mut record = <ValidationRecords<T>>::get(who.clone(), data_id.clone())
```
zees-dev, line 348:
```suggestion
			let mut record = <ValidationRecords<T>>::get(&who, &data_id)
```

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+
+			<ValidationRecords<T>>::insert(who.clone(), data_id.clone(), record);
+
+			Self::deposit_event(Event::ValidationEntryAdded {
+				author: who,
+				id: data_id.to_vec(),
+				checksum,
+			});
+
+			Ok(())
```
zees-dev, line 364: nit:
```suggestion
			<ValidationRecords<T>>::try_mutate(&who, &data_id, |maybe_record| {
            let record = maybe_record.as_mut().ok_or(Error::<T>::RecordNotCreated)?;
            
            record.entries.force_push(ValidationEntry {
                checksum,
                block: <frame_system::Pallet<T>>::block_number(),
            })?;

            Self::deposit_event(Event::ValidationEntryAdded {
                author: who,
                id: data_id.into_vec(),
                checksum,
            });
            
            Ok(())
        })
```

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+			Self::deposit_event(Event::ValidationRecordUpdated {
+				author: who,
+				id: data_id.to_vec(),
+				resolvers: resolvers
+					.map(|resolvers| resolvers.iter().map(|resolver| resolver.to_did()).collect()),
+				data_type: data_type.map(|data_type| data_type.to_vec()),
+				tags: tags.map(|tags| tags.iter().map(|tag| tag.to_vec()).collect()),
+			});
+
+			Ok(())
```
zees-dev, line 407:
```suggestion
			<ValidationRecords<T>>::try_mutate(&who, &data_id, |maybe_record| {
            let record = maybe_record.as_mut().ok_or(Error::<T>::RecordNotCreated)?;

            // Update resolvers if provided
            if let Some(ref new_resolvers) = resolvers {
                Self::validate_sylo_resolvers(new_resolvers)?;
                record.resolvers = new_resolvers.clone();
            }

            // Update data type if provided
            if let Some(ref new_data_type) = data_type {
                record.data_type = new_data_type.clone();
            }

            // Update tags if provided
            if let Some(ref new_tags) = tags {
                record.tags = new_tags.clone();
            }

            Self::deposit_event(Event::ValidationRecordUpdated {
                author: who,
                id: data_id.into_vec(),
                resolvers: resolvers.map(|r| r.iter().map(|resolver| resolver.to_did()).collect()),
                data_type: data_type.map(|dt| dt.into_vec()),
                tags: tags.map(|t| t.iter().map(|tag| tag.into_vec()).collect()),
            });

            Ok(())
        })
```

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+			T::WeightInfo::delete_validation_record()
+		})]
+		pub fn delete_validation_record(
+			origin: OriginFor<T>,
+			data_id: BoundedVec<u8, T::StringLimit>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			ensure!(
+				<ValidationRecords<T>>::get(who.clone(), data_id.clone()).is_some(),
```
zees-dev, line 421:
```suggestion
				<ValidationRecords<T>>::contains_key(&who, &data_id),
```

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+				Error::<T>::RecordNotCreated
+			);
+
+			<ValidationRecords<T>>::remove(who.clone(), data_id.clone());
```
zees-dev, line 425:
```suggestion
			<ValidationRecords<T>>::remove(&who, &data_id);
```

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,459 @@
...
+						<Resolvers<T>>::get(resolver.identifier).is_some(),
+						Error::<T>::ResolverNotRegistered
+					);
+				}
+			}
+
+			Ok(())
+		}
+
+		pub fn payment_asset() -> Option<AssetId> {
```
zees-dev, line 455: seems redundant; can just call the type function directly IMO

pallet/sylo/src/benchmarking.rs
```diff
@@ -0,0 +1,207 @@
...
+		let alice = account::<T>("Alice");
+
+		let mut identifier = BoundedVec::new();
+		for _ in 1..p {
+			identifier.force_push(b'a');
+		}
+
+		let mut service_endpoints = BoundedVec::new();
+		for _ in 1..q {
+			service_endpoints.force_push(bounded_string::<T>("https://service-endpoint.one.two.three"));
```
zees-dev, line 107: probaby push max string size here too; we need upper bound test

pallet/sylo/src/mock.rs
```diff
@@ -0,0 +1,65 @@
...
+	type RuntimeEvent = RuntimeEvent;
+	type MaxResolvers = MaxResolvers;
+	type MaxTags = MaxTags;
+	type MaxEntries = MaxEntries;
+	type MaxServiceEndpoints = MaxServiceEndpoints;
+	type StringLimit = StringLimit;
+	type WeightInfo = ();
+}
+
+#[derive(Default)]
```
zees-dev, line 53: i think we should be able to use the test helpers testext instead od declaring a new/custom one here

pallet/sylo/src/tests.rs
```diff
@@ -0,0 +1,901 @@
...
+					RawOrigin::Signed(bob.clone()).into(),
+					data_id.clone(),
+				),
+				Error::<Test>::RecordNotCreated
+			);
+		});
+	}
+}
+
+fn create_and_register_resolver(
```
zees-dev, line 820: move helper functions to top; i think we follow this approach for most of the tests

pallet/sylo/src/tests.rs
```diff
@@ -0,0 +1,901 @@
...
+
+			assert_noop!(
+				Sylo::unregister_resolver(RawOrigin::Signed(not_controller).into(), identifier,),
+				Error::<Test>::NotController,
+			);
+		});
+	}
+}
+
+mod validation_records {
```
zees-dev, line 205: nit: ideally we try create a `mod` for each extrinsic - which is easier to reason as we can test all flows within the same module.
Examples of this should be in the `crowdsale` or `nft` pallet tests.

zees-dev (commented):
nice; looking mostly good!
Most of the suggestions are purely syntactical improvements and/or improvements to memory.

Atleast 2 major things are still required:
- benchmark weight updates
- e2e tests to validate sylo token payment (NOTE: the refund for any over-payment would be given in the native currency)

Other than that; logic seems sound.

pallet/sylo/src/types.rs
JCSanPedro: I tried something like
```
pub struct ResolverId<Str> {
	pub method: Str,
	pub identifier: Str,
}

pub trait ToDid {
	fn to_did(&self) -> Vec<u8>;
}

impl<T: Get<u32>> ToDid for ResolverId<BoundedVec<u8, T>> {
..
}
```

but didn't feel like it cleaned up the code, as in the lib implementation, I then had to reference the full type like this `ResolverId<BoundedVec<u8, T::StringLimit>>` everywhere

pallet/sylo/src/mock.rs
```diff
@@ -0,0 +1,66 @@
...
+	type ApproveOrigin = EnsureRoot<AccountId>;
+	type MaxResolvers = MaxResolvers;
+	type MaxTags = MaxTags;
+	type MaxEntries = MaxEntries;
+	type MaxServiceEndpoints = MaxServiceEndpoints;
+	type StringLimit = StringLimit;
+	type WeightInfo = ();
+}
+
+// #[derive(Default)]
```
zees-dev, line 54: can remove this comment

pallet/fee-proxy/src/impls.rs
```diff
@@ -182,3 +215,45 @@ where
...
+		_ => false,
+	} {
+		return true;
+	}
+
+	if let Some(pallet_futurepass::Call::proxy_extrinsic { call, .. }) = call.is_sub_type() {
+		return is_sylo_call::<T>(call.as_ref().into_ref());
+	}
+
+	if let Some(call_with_fee_preferences { call, .. }) = call.is_sub_type() {
```
JasonTulp, line 254: Is this part needed as you are returning an error above if the call represents this pattern?

pallet/fee-proxy/src/impls.rs
```diff
@@ -182,3 +215,45 @@ where
...
+		return is_sylo_call::<T>(call.as_ref().into_ref());
+	}
+
```
JasonTulp, line 257: What about sylo calls as an inner call for batch transactions (utility.batch, utility.batch_all etc) or as an inner call for xrpl.transact, doughnuts.transact etc

Check [this doc here](https://www.notion.so/futureverse/Pallets-that-Proxy-Other-Extrinsics-bc5d72b03c5944bfb9a975c507da8c15) which contains a complete list of all extrinsics that contain inner calls, what should the case be for these? Will we need to use Sylo for each? It may be hard to determine, especially for calls such as batch.. I.e. if batch is called alongside other, non Sylo calls do we force all to use Sylo or not enforce any?

pallet/sylo/src/benchmarking.rs
```diff
@@ -0,0 +1,280 @@
...
+
+	return data_id;
+}
+
+benchmarks! {
+	where_clause { where <T as frame_system::Config>::AccountId: From<sp_core::H160> + Into<sp_core::H160> }
+
+	set_payment_asset {}: _(RawOrigin::Root, 24)
+
+	set_sylo_resolver_method {}: _(RawOrigin::Root, bounded_string::<T>("sylo-resolver-method"))
```
JasonTulp, line 100: nit, should add a verify for these 2

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+		#[pallet::call_index(1)]
+		#[pallet::weight({
+			T::WeightInfo::set_sylo_resolver_method()
+		})]
+		pub fn set_sylo_resolver_method(
+			origin: OriginFor<T>,
+			resolver_method: BoundedVec<u8, T::StringLimit>,
+		) -> DispatchResult {
+			T::ApproveOrigin::ensure_origin(origin)?;
+			<SyloResolverMethod<T>>::put(&resolver_method);
```
JasonTulp, line 218: Do we need to verify the validity of the input data? i.e. only utf-8 etc

JasonTulp: Should we allow empty vecs for all extrinsics?

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+		},
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::call_index(0)]
+		#[pallet::weight({
+			T::WeightInfo::set_payment_asset()
+		})]
+		pub fn set_payment_asset(origin: OriginFor<T>, payment_asset: AssetId) -> DispatchResult {
```
JasonTulp, line 202: Although the logic is straight forward, we should have a unit test for this extrinsic

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+			<SyloAssetId<T>>::put(payment_asset);
+			Self::deposit_event(Event::PaymentAssetSet { asset_id: payment_asset });
+			Ok(())
+		}
+
+		#[pallet::call_index(1)]
+		#[pallet::weight({
+			T::WeightInfo::set_sylo_resolver_method()
+		})]
+		pub fn set_sylo_resolver_method(
```
JasonTulp, line 213: Should add an isolated unit test for this extrinsic too

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+		) -> DispatchResult {
+			let reserved_method = <SyloResolverMethod<T>>::get();
+
+			// Ensure any sylo data resolvers are already registered
+			resolvers
+				.iter()
+				.filter(|resolver| resolver.method == reserved_method)
+				.try_for_each(|resolver| -> DispatchResult {
+					ensure!(
+						<Resolvers<T>>::contains_key(&resolver.identifier),
```
JasonTulp, line 462: How many resolvers do we expect to have in prod?

pallet/sylo/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+			origin: OriginFor<T>,
+			data_id: BoundedVec<u8, T::StringLimit>,
+			resolvers: Option<BoundedVec<ResolverId<T::StringLimit>, T::MaxResolvers>>,
+			data_type: Option<BoundedVec<u8, T::StringLimit>>,
+			tags: Option<BoundedVec<BoundedVec<u8, T::StringLimit>, T::MaxTags>>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			<ValidationRecords<T>>::try_mutate(&who, &data_id, |record| -> DispatchResult {
+				let record = record.as_mut().ok_or(Error::<T>::RecordNotCreated)?;
```
JasonTulp, line 394: nit. This error message is a little unclear. Potentially a better name could be `NoRecordExists` or even `NoValidationRecord` to indicate that the failure is due to the lack of the ValidationRecord's existence

JasonTulp (commented):
Nice work John! Very good PR. Logic wise it seems sound, however I have left some comments regarding some edge cases within fee-proxy that may need to be addressed

pallet/fee-proxy/src/impls.rs
```diff
@@ -182,3 +215,45 @@ where
...
+		+ pallet_futurepass::Config
+		+ pallet_sylo::Config,
+	<T as frame_system::Config>::RuntimeCall: IsSubType<crate::Call<T>>,
+	<T as frame_system::Config>::RuntimeCall: IsSubType<pallet_futurepass::Call<T>>,
+	<T as frame_system::Config>::RuntimeCall: IsSubType<pallet_sylo::Call<T>>,
+	<T as Config>::RuntimeCall: IsSubType<pallet_futurepass::Call<T>>,
+	<T as Config>::RuntimeCall: IsSubType<pallet_sylo::Call<T>>,
+	<T as pallet_futurepass::Config>::RuntimeCall: IsSubType<pallet_sylo::Call<T>>,
+{
+	if match call.is_sub_type() {
```
JasonTulp, line 237: You may be able to use the matches! macro here. i.e.
`matches!(call, RuntimeCall::Sylo(..))`
See an example [here](https://github.com/futureversecom/trn-seed/blob/cdc86d5ebc40da7dad4ea037e7b57b7fef4e4d5d/runtime/src/impls.rs#L776)

pallet/sylo/src/lib.rs
JCSanPedro: I think realistically most records would reference 1 resolver

pallet/fee-proxy/src/impls.rs
JCSanPedro: I think I can't use that macro here because the types don't match

pallet/fee-proxy/src/impls.rs
JCSanPedro: I've added more coverage for some proxy calls like Proxy.proxy, xrpl.transaction, and utility.batch.

Some proxy calls are not supported, like the sudo calls, or the scheduler or multisig calls (missing IsSubType trait on their RuntimeCall). For these cases, I think they are okay to just be paid with XRP if for some reason a user ever tries to call a sylo extrinsic in that way.

For doughnut.transact, the [Doughnut Call Validator](https://github.com/futureversecom/trn-seed/blob/main/runtime/src/impls.rs#L934) will prevent a sylo call from being executed in the first place.

For utility.batch, I think it makes sense to allow sylo calls to be batched, but only if all the calls in the batch are sylo extrinsics. That simplifies the fee logic I think.

I tried added more [e2e tests](https://github.com/futureversecom/trn-seed/blob/TRN-808-sylo-pallet/e2e/test/SyloGasCosts.test.ts#L192-L465) to validate the behaviour.

JasonTulp: Nice, reasoning and implementation seem good. Thanks for looking into these

pallet/fee-proxy/src/mock.rs
```diff
@@ -40,6 +40,10 @@ construct_runtime!(
 		Timestamp: pallet_timestamp,
 		Futurepass: pallet_futurepass,
 		FeeControl: pallet_fee_control,
+		Sylo: pallet_sylo,
+		Xrpl: pallet_xrpl,
+		Utility: pallet_utility,
+		Proxy: pallet_proxy,
```
JasonTulp, line 46: Why do you need to add these 3 in the fee proxy mock?

JasonTulp (approved):
Lgtm! Nice work

e2e/common/index.ts
```diff
@@ -623,15 +623,31 @@ export const finalizeTx = (
   extrinsic: SubmittableExtrinsic<"promise">,
   opts?: Partial<SignerOptions>,
 ) => {
-  return new Promise<any[]>((resolve) => {
+  const sendCb =
```
zees-dev, line 626: nice; good to get explicit failure dertails!

e2e/test/Doughnuts.test.ts
```diff
@@ -345,7 +345,7 @@ describe("Doughnuts", () => {
       .transact(call, doughnutHex, nonce, genesis_hash, tip, holderSig)
       .send()
       .catch((err: any) => {
-        console.log(err);
+        console.log("DOUGHNUT ERR", err);
```
zees-dev, line 348: nit: `console.error` this - so it shows up red in cli

e2e/test/SyloGasCosts.test.ts
```diff
@@ -0,0 +1,460 @@
...
+  let node: NodeProcess;
+  let api: ApiPromise;
+  let keyring: Keyring;
+  let alith: KeyringPair;
+  let userPrivateKey: string;
+  let user: KeyringPair;
+  let provider: JsonRpcProvider;
+  let genesisHash: string;
+
+  const FEE_TOKEN_ASSET_ID = 1124;
```
zees-dev, line 42: or you can get `nextAssetId` - other tests seem to do this; the issue with hardcoded asset id's is that the tests dont remain idempotent; i.e. re-runs can result in different output (making them hard to debug)

e2e/test/SyloGasCosts.test.ts
```diff
@@ -0,0 +1,460 @@
...
+    api.tx.syloDataVerification.updateValidationRecord(
+      "data-id",
+      [{ method: "sylo-resolver", identifier: "id-2" }],
+      "data-type-2",
+      ["tag-2"],
+    ),
+    api.tx.syloDataVerification.deleteValidationRecord("data-id"),
+  ];
+
+  it("can submit sylo extrinsic and pay with sylo tokens", async () => {
```
zees-dev, line 111: nit: also worthwhile testing a batchcall tx too.
Unless this is something which is not supported.

e2e/test/SyloGasCosts.test.ts
zees-dev, line 1: this is an awesome set of tests; where most of the edgecases are tested/addressed; nice!

pallet/fee-control/Cargo.toml
```diff
@@ -49,6 +53,8 @@ std = [
 	"frame-system/std",
 	"pallet-assets/std",
 	"pallet-balances/std",
+	"pallet-proxy/std",
+	"pallet-utility/std",
```
zees-dev, line 57: not sure if dev deps need to be in std feature flag.

pallet/fee-proxy/src/impls.rs
```diff
@@ -182,3 +220,131 @@ where
...
+/// fee token.
+///
+/// This will also return an error if the call is an invalid sylo call. A sylo call
+/// can be invalid in the following cases:
+///   - The sylo call has been wrapped in a call_with_fee_preferences call. Sylo
+///     calls should be paid in Sylos only.
+///   - The sylo call is in a batch/batch_all call. In batch calls, if any call is
+///     a sylo call, then all inner calls must be a sylo call. This simplifies the
+///     implementation, preventing a need to process the fee for each individual call.
+fn is_sylo_and_valid_call<T>(
```
zees-dev, line 246: nice!

pallet/fee-proxy/src/impls.rs
```diff
@@ -182,3 +220,131 @@ where
...
+		},
+		Some(pallet_proxy::Call::proxy_announced { call, .. }) => {
+			is_sylo_and_valid_call::<T>(call.as_ref().into_ref())
+		},
+		_ => Ok(false),
+	}?;
+
+	if is_proxy {
+		return Ok(true);
+	}
```
zees-dev, line 301: nit: can remove `is_proxy` check
```suggestion
	let is_proxy = match call.is_sub_type() {
		Some(pallet_proxy::Call::proxy { call, .. }) => {
			return is_sylo_and_valid_call::<T>(call.as_ref().into_ref());
		},
		Some(pallet_proxy::Call::proxy_announced { call, .. }) => {
			return is_sylo_and_valid_call::<T>(call.as_ref().into_ref());
		},
		_ => Ok(false),
	}?;
```

pallet/fee-proxy/src/impls.rs
```diff
@@ -182,3 +220,131 @@ where
...
+		},
+		Some(pallet_utility::Call::as_derivative { call, .. }) => {
+			is_sylo_and_valid_call::<T>(call.as_ref().into_ref())
+		},
+		_ => Ok(false),
+	}?;
+
+	if is_utility {
+		return Ok(true);
+	}
```
zees-dev, line 339: nit: again just return within these; looks cleaner as each of these are distinctive blocks which test inner call.

runtime/src/tests/sylo_fees.rs
zees-dev, line 1: nice tests in runtime

zees-dev (approved):
Nice work; LGTM! 🚀

pallet/fee-control/Cargo.toml
JCSanPedro: Failed to compile without the std feature flag

PR #921, opened 2025-01-30
Fix/migrate nft ownership

JasonTulp:
# Description

- Migrates NFT ownership information into a separate storage map using a multi block migration.

## Context & Background

- The collection Info for an NFT collection is not queryable without using the newly added RPC method. This should make it easier to query that information by separating out the ownership info for an NFT collection

## Notes & Additional Information

- Everywhere that accesses the NFT collection Info or ownership info must return an error while the migration is ongoing, this is handled largely by the impl of NFTExt within the NFT pallet

## Related Issues

- https://futureverse.atlassian.net/browse/TRN-240?atlOrigin=eyJpIjoiM2MyYWY4OWU1ZTFlNGU4YjhjNjNjNzA1ZDZlMjZiY2QiLCJwIjoiaiJ9
- 
---

# Release Notes

## Type of Change

- [x] Runtime Changes
- [ ] Client Changes

#### Added

- Nft: OwnershipInfo

#### Changed

- Nft: CollectionInfo

#### Added

- Nft Multiblock migration: migrate token ownership

#### Removed

-

---


pallet/nft/src/impls.rs
```diff
@@ -49,19 +49,25 @@ impl<T: Config> Pallet<T> {
 
 	/// Return whether the collection exists or not
 	pub fn collection_exists(collection_id: CollectionUuid) -> bool {
+		if T::Migrator::ensure_migrated().is_err() {
+			return false;
+		}
```
zees-dev, line 54: we are returning potentially false info here; instead we should simply return a generic migration error.
A caller (and respective downstream SDKs/libraries/UI's may display incorrect info).

pallet/nft/src/impls.rs
```diff
@@ -49,19 +49,25 @@ impl<T: Config> Pallet<T> {
...
 	}
 
 	/// Returns number of tokens owned by an account in a collection
 	/// Used by the ERC721 precompile for balance_of
 	pub fn token_balance_of(who: &T::AccountId, collection_id: CollectionUuid) -> TokenCount {
-		match <CollectionInfo<T>>::get(collection_id) {
-			Some(collection_info) => {
-				let serial_numbers: Vec<SerialNumber> = collection_info
+		if T::Migrator::ensure_migrated().is_err() {
+			return 0;
```
zees-dev, line 62: same applies for this function (and all the rest)

pallet/nft/src/impls.rs
zees-dev: It may just be easier to move this statement to the first line of the respective extrinsic (in the lib file)

zees-dev: ignore comments above; i've realized other pallets can call these impls; hence why they are here. Makes sense.

zees-dev (approved):
LGTM 👍 
Just need CI to pass.

pallet/nft/src/impls.rs
```diff
@@ -307,9 +313,11 @@ impl<T: Config> Pallet<T> {
 			Error::<T>::TokenLimitExceeded
 		);
 
-		new_collection_info
+		let mut ownership_info = OwnershipInfo::<T>::get(collection_id).unwrap_or_default();
+		ownership_info
 			.add_user_tokens(token_owner, serial_numbers.clone())
-			.map_err(Error::<T>::from)?;
+			.map_err(|e| Error::<T>::from(e))?;
+		<OwnershipInfo<T>>::insert(collection_id, ownership_info);
```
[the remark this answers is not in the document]
JasonTulp: I think it's better to insert at the end in case the add_user_tokens fails, then the storage will not need to be reverted

zees-dev (approved):
LGTM 👍

pallet/nft/src/lib.rs
```diff
@@ -78,9 +78,10 @@ pub mod pallet {
 	use super::{DispatchResult, *};
 	use frame_support::pallet_prelude::*;
 	use frame_system::pallet_prelude::*;
+	use seed_pallet_common::Migrator;
 
 	/// The current storage version.
-	const STORAGE_VERSION: StorageVersion = StorageVersion::new(5);
+	const STORAGE_VERSION: StorageVersion = StorageVersion::new(8);
```
surangap, line 84: How come it jumped from 5 to 8? missed to update?

surangap (commented):
lgtm. left some comments. also we might need some bench run for nft pallet now the storage structures are different.

runtime/src/migrations/nft_multi.rs
```diff
@@ -0,0 +1,413 @@
...
+	old: old::CollectionInformation<T::AccountId, T::MaxTokensPerCollection, T::StringLimit>,
+) -> Result<
+	(
+		CollectionInformation<T::AccountId, T::StringLimit>,
+		TokenOwnership<T::AccountId, T::MaxTokensPerCollection>,
+	),
+	&'static str,
+> {
+	// Construct ownership info out of old ownership info
+	let token_ownership_old = old.owned_tokens;
```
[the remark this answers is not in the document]
JasonTulp: It's roughly ~0.3MB at the worst case for one collection.
We will reduce the risk by performing one collection per block, by setting BlockLimit to 1 in migration pallet
There are 61 collections on mainnet that need migrating, only 3 of them are this size so shouldn't be an issue and will take only 4 minutes to complete

pallet/nft/src/lib.rs
JasonTulp: Yeah that's my bad.. Forgot to update the default StorageVersion

JasonTulp: It's on-chain version is 7

surangap (approved):
need CI bench runs for other affected pallets as well.

evm-precompiles/erc721/src/lib.rs
```diff
@@ -1257,16 +1256,13 @@ where
 		let owner: H160 = owner.into();
 
 		handle.record_cost(RuntimeHelper::<Runtime>::db_read_gas_cost())?;
```
JCSanPedro, line 1258: Does calling `iter_prefix` on a `StorageNMap` mean that we read the db for each value we get?

So then would it make sense to record the gas cost for each item?

JasonTulp: Yeah good point, would be better to do it like this where we record the cost right after reading the next value in the iter to prevent issues when the iterator becomes too large
https://github.com/futureversecom/trn-seed/pull/921/commits/c49560da6ea09195690dcc1df8677c05b01567fc

runtime/src/migrations/nft_multi.rs
```diff
@@ -0,0 +1,683 @@
...
+) -> (bool, TokenCount) {
+	log::debug!(target: LOG_TARGET, "🦆 Migrating collection_id {}", collection_id);
+	// Is the collection info fully migrated
+	let mut completed = false;
+	let mut migrated_token_count: TokenCount = 0;
+
+	// For simplicity, we will only migrate max 50 tokens from 1 account at a time
+	if let Some(mut ownership) = old.owned_tokens.pop() {
+		let mut serial_numbers = ownership.owned_serials.clone();
+		// take at max 10 from serial_numbers
```
felixshu, line 121: // take at max MAX_TOKENS_PER_STEP from serial_numbers

runtime/src/migrations/nft_multi.rs
```diff
@@ -0,0 +1,683 @@
...
+			let token_info = TokenInformation {
+				owner: ownership.owner.clone(),
+				lock_status: token_locks,
+				utility_flags: token_utility_flags,
+			};
+			pallet_nft::TokenInfo::<T>::insert(collection_id, serial_number, token_info);
+		}
+
+		// Update OwnedTokens with the migrated serials
+		let _ = pallet_nft::OwnedTokens::<T>::try_mutate(
```
felixshu, line 140: we ignored the error in owned tokens update

if let Err(e) = pallet_nft::OwnedTokens::<T>::try_mutate(
    &ownership.owner,
    collection_id,
    |owned_serials| -> DispatchResult {
        // ...
        Ok(())
    },
) {
    log::error!(target: LOG_TARGET, "🦆 Error updating owned tokens: {:?}", e);
    // Consider how to handle this error - maybe mark the migration as failed?
}

JasonTulp: The try_mutate will never error as we always return Ok(())
In both places where it could have errored, we are using force_push and truncate_from which do not return a Result. These are both safe as the BoundedVec upper bound will not change between the old and new datasets
so it is safe to ignore the result altogether

pallet/sft/src/lib.rs
```diff
@@ -134,18 +135,20 @@ pub mod pallet {
 
 	// Map from a collection id to a collection's pending issuances
 	#[pallet::storage]
-	pub type PendingIssuances<T: Config> = StorageMap<
+	pub type PendingIssuances<T: Config> = StorageNMap<
```
zees-dev, line 138: Interesing use of new storage data structure :+1

PR #926, opened 2025-02-17
Pass attribution - Futurepass integration

zees-dev:
# Description

- Implemented `createFuturepassWithPartner` extrinsic
  - Futurepass as FP provider to create an FP and assign to a partner in 1 extrinsic - permissionlessly
- Removed the functionality for a partner to de-register and re-register
- Added and updated unit tests, e2e tests and benchmarks

---

# Release Notes

## Type of Change

- [x] Runtime Changes
- [ ] Client Changes

#### Added

- **PartnerAttribution: `create_futurepass_with_partner`**

#### Changed

- **PartnerAttribution: `update_partner_account`**; removed functionality to de-register partner

---


pallet/partner-attribution/src/mock.rs
```diff
@@ -54,5 +54,6 @@ impl crate::Config for Test {
 	type RuntimeEvent = RuntimeEvent;
 	type ApproveOrigin = EnsureRoot<Self::AccountId>;
 	type EnsureFuturepass = EnsureAny;
+	type FuturepassCreator = ();
```
JasonTulp, line 57: It may be better to include the Futurepass pallet as the implementor of this for the tests, that way we can test the flow where create_futurepass fails. i.e. if Holders already exists

pallet/partner-attribution/src/lib.rs
```diff
@@ -210,7 +237,7 @@ pub mod pallet {
 		///
 		/// Parameters:
 		/// - `partner_id`: The partner id to attribute the account to.
-		#[pallet::call_index(2)]
+		#[pallet::call_index(3)]
```
JasonTulp, line 240: I think we should keep the call_indexes the same and add the new extrinsic as index 4, that way this will ensure that this change won't cause breaking changes downstream. (Although unlikely)

pallet/partner-attribution/src/lib.rs
```diff
@@ -166,39 +169,63 @@ pub mod pallet {
...
-			partner_account: Option<T::AccountId>,
+			partner_account: T::AccountId,
 		) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
 			// Get partner and verify ownership in one read
 			let partner = Partners::<T>::get(partner_id).ok_or(Error::<T>::PartnerNotFound)?;
 			ensure!(partner.owner == who, Error::<T>::Unauthorized);
 
-			match partner_account {
```
JasonTulp, line 183: I understand it is probably a business requirement, but am interested to hear the reasoning behind restricting partner removal functionality

pallet/common/src/lib.rs
```diff
@@ -651,6 +651,41 @@ pub trait SFTExt {
...
+		// Add the owner's account bytes
+		input.extend_from_slice(&owner.encode());
+
+		// Hash the input to get a deterministic address
+		let hash = sp_io::hashing::blake2_256(&input);
+		let address = H160::from_slice(&hash[0..20]);
+
+		Ok(address.into())
+	}
+}
```
JasonTulp, line 687: nit, if wanting to create some mock behaviour for this trait, it should be done within the mock file under a MockFuturepassProvider instead of as an implementation on the unit type

If wanting to create a unit type impl, this should return an error to prevent accidental behaviour

Personally I think this impl should be removed in favour of using the real Futurepass pallet in the tests to ensure the Futurepass account is created correctly with appropriate values

JasonTulp (approved):
lgtm! Few minor changes to do with the unit tests

pallet/partner-attribution/src/mock.rs
zees-dev: Can do this but it creates a dependency and increases build times.
Since futurepass pallet depends on runtime - hence building the tests just becomes too large of a task :/

pallet/common/src/lib.rs
zees-dev: Have mixed opinions on this since - again you create a dependency on the futurepass pallet.
And again the FP pallet has dependencies on other things - mainly the runtime atm.

I have seen many times in the core substrate libs where the unit type is used alongside the trait declaration - I think this is good design as in most cases consumers just want to plug some default impl.
We are providing the same here - in most cases, if users want an impl which wants to hook into storage - then they will need an explicit import of the FP pallet itself as the provider.

The current impl does give a FP address - just without a storage.

pallet/partner-attribution/src/lib.rs
zees-dev: https://github.com/futureversecom/trn-seed/pull/926/commits/d93d4f0b7f616dd13cd1a7c865fa6354780d92c6

pallet/common/src/lib.rs
JasonTulp: In this case we should move this to mock. I've seen the unit type impls and have used them myself, however all the impls on the unit type are intended to be no-op, whereas this impl creates some random account from the owner address which.
Though unlikely, if this unit type is accidentally used incorrectly, it could result in unwanted behaviour

surangap: We have similar implementations in our code where the unit type having mocked-like implementation.

surangap (approved):
lgtm

pallet/partner-attribution/src/mock.rs
JasonTulp: Right, I understand that. The main reason for me bringing this up is because the `create_futurepass_when_already_attributed_fails` test would actually fail with a different error if ran with the real Futurepass pallet, hence this test is not 100% accurate. (It would fail with `pallet_futurepass::Error::AccountAlreadyRegistered`)

Up to you which approach to take, is not a biggie either way :slightly_smiling_face:

pallet/common/src/lib.rs
JasonTulp: No worries, it's not a biggie. Don't want to hold up release for this

JasonTulp (approved):

pallet/common/src/lib.rs
zees-dev: https://github.com/futureversecom/trn-seed/pull/926/commits/dc5112e1d331092d9782fe9cbfe618e1b1325ac5

pallet/partner-attribution/src/mock.rs
zees-dev: https://github.com/futureversecom/trn-seed/pull/926/commits/dc5112e1d331092d9782fe9cbfe618e1b1325ac5

PR #929, opened 2025-02-25
Pass attribution partner fee accumulation via fee-proxy

zees-dev:
# Description

- Updated the `FeeProxy` pallet to allow partners to accumulate fees via the `PartnerAttribution` pallet - if the partner has a fee percentage set (by an admin account).
- Added unit tests + mocks
- Added E2E test to validate partner fee accumulation on FP transaction

---

# Release Notes

## Key Changes
- Updated the `FeeProxy` pallet to allow partners to accumulate fees via the `PartnerAttribution` pallet - if the partner has a fee percentage set (by an admin account).

## Type of Change

- [x] Runtime Changes
- [ ] Client Changes

---

### DIFF
--- a/pallet/fee-proxy/src/impls.rs
+++ b/pallet/fee-proxy/src/impls.rs
@@ -32,6 +32,7 @@ where
 		+ pallet_assets_ext::Config
 		+ pallet_futurepass::Config
 		+ pallet_sylo_data_verification::Config
+		+ pallet_partner_attribution::Config
 		+ pallet_proxy::Config
 		+ pallet_utility::Config
 		+ pallet_xrpl::Config,
@@ -81,6 +82,9 @@ where
 				who = futurepass;
 			}
 		}
+
+		// Attribute the fee to the partner if the caller is attributed to a partner
+		attribute_fee_to_partner::<T>(who, fee.into())?;
 
 		let do_fee_swap = |who: &T::AccountId,
 		                   payment_asset: &AssetId,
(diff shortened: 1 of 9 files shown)


pallet/fee-proxy/src/impls.rs
```diff
@@ -221,6 +225,28 @@ where
...
+	fee: Balance,
+) -> Result<(), TransactionValidityError>
+where
+	T: Config + pallet_partner_attribution::Config,
+{
+	if let Some(partner_id) = pallet_partner_attribution::Attributions::<T>::get(who) {
+		pallet_partner_attribution::Partners::<T>::try_mutate(partner_id, |maybe_partner| {
+			if let Some(partner) = maybe_partner {
+				if partner.fee_percentage.is_some() {
+					partner.accumulated_fees = partner.accumulated_fees.saturating_add(fee);
```
JasonTulp, line 241: Might be an idea to add a checked_add and then error

pallet/fee-proxy/src/tests.rs
```diff
@@ -440,3 +440,69 @@ mod calculate_total_gas {
...
+					&call,
+					&DispatchInfo::default(),
+					fee.into(),
+					0u32.into(),
+				)
+				.expect("Fee withdrawal should work");
+
+				// Verify partner's accumulated fees increased
+				let updated_partner = Partners::<Test>::get(partner_id).unwrap();
+				assert!(
```
JasonTulp, line 496: Why check if it's above 0 if you're checking the exact value after this?

JasonTulp (approved):
lgtm!

pallet/fee-proxy/src/tests.rs
zees-dev: https://github.com/futureversecom/trn-seed/pull/929/commits/eab6f9e4f0ec09dd854c25f4acaa00852a978d76

pallet/fee-proxy/src/impls.rs
zees-dev: If the partner has maxed out on accumulated fees (unlikely); then all transactions from attributed accounts will start to fail; we probably dont want this - hence i think it makes sense to retain saturating add?

surangap (approved):
lgtm

pallet/fee-proxy/src/impls.rs
JasonTulp: Hmm good point, sounds good. Would be better to just pay them less than stall their attributed accounts

PR #937, opened 2025-04-07
Sm 462 sylo permissions pallet

JCSanPedro:
# Description

RFC - https://www.notion.so/futureverse/RFC-Sylo-Data-Permissions-Protocol-1b20cb4dab3d8037ab64f9eef1bf9dea

The **Sylo Data Permissions** pallet is used to store permission records that allow access control mechanisms for 
Storage Sylos. The pallet has two primary users:
- `Data Authors`/`Grantors` - These are parties that store off-chain data in Storage Sylos and use the `Sylo Data Verification` pallet for storing validation records. `Data Authors` can grant other accounts permissions to access, modify, or distribute their data.
- `Storage Sylos` - Sylos will use read from the pallet to determine if an account that is accessing their APIs to view or modify stored data have the correct permissions to do so.

This pallet integrates with the **Sylo Data Verification**, and each permission record is generally associated to an onchain validation record.

There are three types of permission records stored onchain.

- **Permission Records** - The default permission record grants an account a permission for a single accompanying data validation record that is owned by the data author/grantor.
- **Tagged Permission Records** - A tagged permission record has a `tags` field. This permission will apply to any data validation records owned by the grantor, that also share any of the tags in the permission record.
- **Permission Reference** - A record that references another piece of data that is stored off-chain. This off-chain data is intended to be a JSON value which states a large number of permissions. The off-chain data is not validated by the pallet directly, but there must exist an on-chain validation record for it.

There are also three levels of data permissions: 
- `VIEW` - Allows an account to fetch/download the data
- `MODIFY` - Allows an account to make updates to the off-chain data. **Also** allows an account to update the associated on-chain validation record via [add_validation_record_entry](https://github.com/futureversecom/trn-seed/blob/SM-462-sylo-permissions-pallet/pallet/sylo-data-verification/src/lib.rs#L383).
- `DISTRIBUTE` - Allows an account to create additional permission records on behalf of the data author

## Pallet RPC

An additional RPC has been added for this pallet. The RPC call `get_permissions` makes it easier for a client to determine what permissions an account has for multiple data items. As there are 3 types of permission records stored on-chain, a client would typically need to query 3 seperate storages for each data item. This RPC call makes it so that a client only needs to send one query to retrieve the permissions.

The response includes a `permission_reference` record. This is only filled if the account has been granted a permission reference. The value is an object which includes the id of the offchain permission record, and also a list of resolvers with their endpoints.

```
pub struct PermissionReferenceRecord {
	pub permission_record_id: String,
	pub resolvers: Vec<(String, Vec<String>)>, // list of resolvers + endpoints
}

pub struct HasPermissionQueryResult {
	pub onchain: bool,
	pub permission_reference: Option<PermissionReferenceRecord>,
}

#[method(name = "has_permission_query")]
fn has_permission_query(
	&self,
	data_author: AccountId,
	grantee: AccountId,
	data_id: String,
	permission: DataPermission,
) -> RpcResult<Result<HasPermissionQueryResult, DispatchError>>;
```

---

# Release Notes

## Type of Change

- [x] Runtime Changes
- [x] Client Changes

#### Added

- **NextPermissionRecordId** - Counter for permission record ids
- **PermissionRecords** - Holds a vector of on-chain permission records under (data_author, grantee, data_id)
- **TaggedPermissionRecords** - Holds a vector of tagged permission records (data_author, grantee)
- **PermissionReference** - Holds a permission reference record keyed by (data_author, grantee)

#### Added

- **grant_data_permissions** - Grant another account data permissions
- **revoke_data_permissions** - Revokes previously granted data permissions
- **grant_tagged_permissions** - Grants another account tagged data permissions
- **revoke_tagged_permissions** - Revokes previously granted tagged data permissions
- **grant_permission_reference** - Sets a permission reference record for another account
- **revoke_permission_reference** - Removes a previously granted permission reference

#### Added

- **Sylo Data Permissions: DataPermissionGranted**
- **Sylo Data Permissions: DataPermissionRevoked**
- **Sylo Data Permissions: TaggedDataPermissionGranted**
- **Sylo Data Permissions: TaggedDataPermissionRevoked**
- **Sylo Data Permissions: PermissionReferenceGranted**
- **Sylo Data Permissions: PermissionReferenceRevoked**

#### Added

- **Sylo Data Permissions: DataRecordDoesNotExist**
- **Sylo Data Permissions: IrrevocableCannotBeExpirable**
- **Sylo Data Permissions: ExceededMaxPermissions**
- **Sylo Data Permissions: MissingDistributePermission**
- **Sylo Data Permissions: CannotGrantDistributePermission**
- **Sylo Data Permissions: CannotGrantIrrevocablePermission**
- **Sylo Data Permissions: PermissionIrrevocable**
- **Sylo Data Permissions: NotPermissionGrantor**
- **Sylo Data Permissions: PermissionNotFound**
- **Sylo Data Permissions: MissingValidationRecord**
- **Sylo Data Permissions: InvalidString**

---


pallet/sylo-data-permissions/src/lib.rs
```diff
@@ -0,0 +1,690 @@
...
+			permission: DataPermission,
+			expiry: Option<BlockNumberFor<T>>,
+			irrevocable: bool,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			let block = <frame_system::Pallet<T>>::block_number();
+
+			let permission_record =
+				PermissionRecord { grantor: who.clone(), permission, block, expiry, irrevocable };
```
felixshu, line 249: we may need to verify the expiry ? the exipry could be none, but when user set the number, it should bigger than block?

If expiry <= current_block, the permission is considered expired. This means:
1. If a caller sets expiry to the current block, the permission would be immediately expired
2. If a caller sets expiry to a past block, the permission would never work

pallet/sylo-data-permissions/src/lib.rs
```diff
@@ -0,0 +1,690 @@
...
+		{
+			let tagged_permissions = <TaggedPermissionRecords<T>>::get(data_author, grantee);
+
+			return tagged_permissions.iter().any(|(_, record)| {
+				// check for expiry
+				if let Some(expiry) = record.expiry {
+					if expiry <= block {
+						return false;
+					}
+				}
```
felixshu, line 569: permission expires at the start of its expiry block? Using < would allow it to remain valid through its expiry block. or consider minimum expiration gaps.

pallet/sylo-data-permissions/src/lib.rs
```diff
@@ -0,0 +1,690 @@
...
+				T::SyloDataVerificationProvider::get_validation_record(&who, &permission_record_id)
+					.is_some(),
+				Error::<T>::MissingValidationRecord,
+			);
+
+			<PermissionReferences<T>>::insert(
+				&who,
+				&grantee,
+				PermissionReference { permission_record_id: permission_record_id.clone() },
+			);
```
felixshu, line 481: `PermissionReferences<T>` is defined as a `StorageDoubleMap` at uses `(grantor, grantee)` as keys and stores a single `PermissionReference` value.

When a new permission reference is granted, the function uses `.insert()` which unconditionally overwrites any existing permission reference for the same (grantor, grantee) pair.

so:
  1. There's no check before insertion to see if an existing permission reference already exists.
  2. The function provides no warning or notification that an existing reference is being replaced.

This creates several potential issues:

- Silent replacement of existing permission references
- No history of changes (each reference simply replaces the previous one)
- Users may be unaware that their previous permission reference was overwritten
- No way to detect accidental overwrites

pallet/sylo-data-permissions/rpc/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+use sp_blockchain::HeaderBackend;
+use sp_runtime::traits::Block as BlockT;
+use sp_runtime::DispatchError;
+
+use pallet_sylo_data_permissions::GetPermissionsResult;
+pub use pallet_sylo_data_permissions_rpc_runtime_api::{
+	self as runtime_api, SyloDataPermissionsApi as SyloDataPermissionsRuntimeApi,
+};
+use seed_pallet_common::sylo::DataPermission;
+
```
felixshu, line 37: check the unused import here

pallet/sylo-data-permissions/src/lib.rs
```diff
@@ -0,0 +1,711 @@
...
+				// check for expiry
+				if let Some(expiry) = record.expiry {
+					if expiry < block {
+						return false;
+					}
+				}
+
+				return is_sufficient_permission(record.permission);
+			})
+			.is_some();
```
JasonTulp, line 572: Can use .any here instead of .find().is_some()

pallet/sylo-data-permissions/src/lib.rs
```diff
@@ -0,0 +1,711 @@
...
+					if expiry < block {
+						return false;
+					}
+				}
+
+				// check if any of the tags in the permission record
+				// matches any of the tags in data record
+				return record.tags.iter().any(|permission_tag| {
+					verification_record.tags.iter().any(|record_tag| permission_tag == record_tag)
+				}) && is_sufficient_permission(record.permission);
```
JasonTulp, line 596: nit, can slightly optimise by putting `is_sufficient_permission` at the start of the && operation

pallet/sylo-data-verification/src/lib.rs
```diff
@@ -368,34 +374,51 @@ pub mod pallet {
...
-				let record = record.as_mut().ok_or(Error::<T>::NoValidationRecord)?;
+			ensure!(
+				who == data_author
+					|| T::SyloDataPermissionsProvider::has_permission(
+						&data_author,
+						&data_id,
+						&who,
+						DataPermission::MODIFY
+					),
+				Error::<T>::MissingModifyPermission
```
JasonTulp, line 399: Should update the benchmark for this extrinsic to traverse the worst case for the added complexity of the has_permission check. 
has_permission adds 3 reads that should be included in the weight

pallet/sylo-data-permissions/src/lib.rs
```diff
@@ -0,0 +1,711 @@
...
+			})
+			.is_some();
+
+		if has_direct_permission {
+			return true;
+		}
+
+		// check for tagged permissions
+		if let Some(verification_record) =
+			T::SyloDataVerificationProvider::get_validation_record(&data_author, &data_id)
```
JasonTulp, line 580: It may be better to pass the validation record into this function, rather than fetching from the provider trait. In the verification pallet you are calling this has_permission method then mutating ValidationRecords after which is adding an extra read

pallet/sylo-data-verification/src/lib.rs
```diff
@@ -505,3 +528,69 @@ pub mod pallet {
...
+								.map(|resolver| resolver.service_endpoints)
+								.unwrap_or(BoundedVec::new()),
+						);
+					} else {
+						return (resolver_id, BoundedVec::new());
+					}
+				})
+				.collect::<Vec<_>>(),
+		)
+		.unwrap() // safe to unwrap here
```
JasonTulp, line 594: If the length of the vec is guaranteed to fit within the bounds you can use truncate_from instead of try_from().unwrap()

JasonTulp (approved):
Nice work! Super clear PR. Nice to see the thorough test coverage including RPC test! :) Left a few comments, main one is the benchmark

pallet/sylo-data-permissions/src/lib.rs
```diff
@@ -0,0 +1,711 @@
...
+				PermissionRecord { grantor: who.clone(), permission, block, expiry, irrevocable };
+
+			if irrevocable {
+				ensure!(expiry.is_none(), Error::<T>::IrrevocableCannotBeExpirable);
+			}
+
+			if let Some(expiry) = expiry {
+				// ensure valid expiry
+				ensure!(expiry > block, Error::<T>::InvalidExpiry);
+			}
```
JasonTulp, line 262: Currently the expired entries will remain in storage with no mechanism to clear them. In a future iteration it may be an idea to implement an automated cleaning system that will remove these entries from storage. This can be done through on_idle by adding the next_id into a separate map, keyed by the block_number. Then every block this pallet can check the block number to see if any entries need clearing. 
Worth thinking about

JCSanPedro: Interesting change. I think though we would to retain expired records so that a user can know if  one of their previously granted permissions became expired, and would allow them to reinstate it.

For example if we had a front-end experience that allowed users to grant permissions, the front end could show both active and expired records. However if the records were deleted immediately, then a user could miss that the record expired (unless we had some other indexing service to retain the data).

JasonTulp: Understandbale, that makes sense and I agree it could be useful to see expired records from a UX perspective.  In this scenario however, if a record is "reinstated" would it overwrite the expired entry or create a brand new permissions entry? 
From a storage perspective, it is best to clear data that is not being used, so maybe a solution could look like:
- expired permission entries are removed automatically after a fixed time period after expiration (i.e. 1 month) 
- Expired permission entries within this time period can be reinstated which would extend the expiry date.

What are your thoughts? 
Either way it doesn't need to be in this iteration

JasonTulp (approved):
lgtm!

pallet/sylo-data-permissions/src/lib.rs
```diff
@@ -0,0 +1,889 @@
...
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		/// Check and close all expired listings
+		fn on_initialize(now: BlockNumberFor<T>) -> Weight {
+			let (removed_permissions, removed_tagged_permissions) =
+				Self::do_remove_expired_permissions(now);
+
+			<T as Config>::WeightInfo::on_initialize(
```
JasonTulp, line 296: Nice use of a benchmark here!

pallet/sylo-data-permissions/src/lib.rs
```diff
@@ -0,0 +1,889 @@
...
+		Ok(())
+	}
+
+	/// Removes all data permissions and tagged permissions that have expired.
+	/// Returns the number of permissions removed for both sets.
+	pub fn do_remove_expired_permissions(now: BlockNumberFor<T>) -> (u32, u32) {
+		let mut removed_permissions = 0;
+
+		<ExpiringPermissionRecords<T>>::mutate(now, |records| {
+			for (data_author, grantee, data_id, permission_id) in records.iter() {
```
JasonTulp, line 847: Instead of iter here with a .clear at the end, you can use records.drain(..)
This will remove the items from the list and return them as an iterator
This will also give ownership of the data so you can remove all clones in the event and pass in references to the mutate

pallet/sylo-data-permissions/src/lib.rs
```diff
@@ -0,0 +1,889 @@
...
+		ValueQuery,
+	>;
+
+	#[pallet::storage]
+	pub type ExpiringPermissionRecords<T: Config> = StorageMap<
+		_,
+		Twox64Concat,
+		BlockNumberFor<T>,
+		BoundedVec<
+			(T::AccountId, T::AccountId, DataId<T::StringLimit>, u32),
```
JasonTulp, line 148: It's a little unclear what these tuple values are without looking through the code, do you mind adding a comment to the storage items to show what the key comprises of?

JasonTulp (approved):
Thanks for adding the storage clearing mechanism! And nice job on the clean benchmark based weight calculation. This is the way we should be doing all of these types of automated systems. Left a few minor comments but nothing critical. Nice one

felixshu (approved):

pallet/sylo-data-permissions/src/lib.rs
JCSanPedro: I think `drain` isn't actually suitable here because the type of storage is `StorageMap<BlockNumber, BoundedVec<..>>`. If we use `drain` there would only be one element in the iterable (which is the bounded vec), and we would still need to call `.iter()` on the bounded vec value.

I updated the implementation to use `take()` on the storage value instead of `mutate()`, which I think achieves the same level of clarity as to what you suggested.

PR #958, opened 2025-07-07
Add Additional Data field to NFT and SFT

JasonTulp:
# Description

- Adds an additional data field to NFT and SFT, allowing collection owners to assign extra data to their tokens

## Notes & Additional Information

- Adds 2 extra extrinsics to each pallet, one which lets the user add data to existing tokens, one which allows minting of new tokens with this field

## Related Issues

- https://futureverse.atlassian.net/browse/TRN-1007?atlOrigin=eyJpIjoiY2ZmNTk0NzI3MTRlNGM5Zjk1NzI1ZjAxY2Q5ZGMwNTkiLCJwIjoiaiJ9

---

# Release Notes

## Type of Change

- [x] Runtime Changes
- [ ] Client Changes

#### Added

- _e.g. **XRPLBridge: PaymentDelay, changed from value to map keyed by asset_id**_

#### Changed

-

#### Removed

-

#### Added

-

#### Changed

- _e.g. **XRPLBridge: set_payment_delay, added asset_id**_

#### Removed

-

#### Added

- _e.g. **XRPL-Bridge: XrplAssetMapSet**_

#### Changed

- _e.g. **XRPLBridge: PaymentDelaySet, added asset_id parameter**_

#### Removed

-

#### Added

- _e.g. **Marketplace: NoToken**_

#### Changed

-

#### Removed

-

#### Added

- _e.g. **XRPLBridge: Migrate PaymentDelay to include asset_id as key**_

#### Removed

-

---


runtime/src/lib.rs
```diff
@@ -459,6 +459,7 @@ parameter_types! {
 	pub const MaxTokensPerCollection: u32 = 1_000_000;
 	pub const MintLimit: u32 = 1_000;
 	pub const TransferLimit: u32 = 1_000;
+	pub const NftAdditionalDataLength: u32 = 32;
```
JCSanPedro, line 462: This limit seems small? I guess we are trying to be conservative here, but 32 byte limit means I would essentially just be able to use a single tag for the additional metadata, but I can imagine it could be nice to use a larger serialized json value as the additional metadata.

JCSanPedro (approved):
lgtm 👍

runtime/src/lib.rs
JCSanPedro: I think 100 seems reasonable. I imagine 32 is what is needed for the actual requirements, but a bigger limit would make this functionality more applicable for a wider use case

pallet/nft/src/lib.rs
```diff
@@ -904,6 +873,63 @@ pub mod pallet {
...
+			ensure!(collection_info.is_collection_owner(&who), Error::<T>::NotCollectionOwner);
+			ensure!(collection_info.token_exists(token_id.1), Error::<T>::NoToken);
+			Self::do_set_additional_data(token_id, additional_data)?;
+			Ok(())
+		}
+
+		/// Mint a token alongside some additional data
+		#[pallet::call_index(17)]
+		#[pallet::weight(T::WeightInfo::mint_with_additional_data())]
+		pub fn mint_with_additional_data(
```
surangap, line 900:
```suggestion
                #[transactional]
		pub fn mint_with_additional_data(
```

surangap (approved):
lgtm

